PR #40, opened 2020-03-27
Added basic support for i18n

sgirgin:
Currently, this covers HTML templates and not JS files.


sgirgin:
> awesome! This is perfect.
> There might be some stuff in js, but for sure there is one in time_ago() that returns a string. Maybe it should return a unit and a number instead and format in HTML. you wanna look at this next ?

Sure, I'll take a look at that next (also to strings in JS).

PR #44, opened 2020-03-27
Dulacarnold db

dulacarnold:


icubam/db/sqlite.py
```diff
@@ -167,13 +167,6 @@ def get_bedcount(self, icu_ids=None, max_ts=None):
     query += """ GROUP BY icu_id"""
     return pd.read_sql_query(query, self._conn)
 
-  def get_bedcount_by_icu_id(self, icu_id):
```
olivierteboul: this is used in update.py. please make sure to fix it

PR #45, opened 2020-03-27
Migrated sqlite3 to sqlalchemy.

sgirgin:
We are still using SQLite do I kept the class the same. We can later change it.

### DIFF
--- a/icubam/db/sqlite.py
+++ b/icubam/db/sqlite.py
@@ -1,61 +1,76 @@
 """SQLite storage backend wrapper."""
 import logging
 import os
-import sqlite3
 import time
 import pandas as pd
+from sqlalchemy import create_engine
+from sqlalchemy import Table, Column, MetaData, ForeignKey, UniqueConstraint
+from sqlalchemy import Float, Integer, String
 
 
 class SQLiteDB:
   """Wraps SQLite DB for bed counts."""
 
   def __init__(self, db_path: str):
-    """Given a token file and a sheet id, loads the sheet to be queried."""
-    self._db_path = db_path
-
-    if os.path.exists(db_path):
-      self._conn = sqlite3.connect(db_path)
-    else:
-      self._conn = sqlite3.connect(db_path)
-      self._create_table()
-
-  def _create_table(self):
-    self._conn.execute(
-      """CREATE TABLE icus
-                          (icu_id INTEGER NOT NULL PRIMARY KEY,
-                           icu_name TEXT UNIQUE, dept TEXT, city TEXT, lat REAL,
-                           long REAL, telephone TEXT )"""
+    """Initializes the database."""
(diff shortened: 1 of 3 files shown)


icubam/db/sqlite.py
```diff
@@ -12,52 +14,65 @@ class SQLiteDB:
...
-      """CREATE TABLE icus
-                          (icu_id INTEGER NOT NULL PRIMARY KEY,
-                           icu_name TEXT UNIQUE, dept TEXT, city TEXT, lat REAL,
-                           long REAL, telephone TEXT )"""
+    """Initializes the database."""
+    self._metadata = MetaData()
+    self._icus = Table(
+        'icus',
+        self._metadata,
+        Column('icu_id', Integer, primary_key=True),
```
dulacarnold: Can we enforce relational constraints on this column: https://www.sqlite.org/foreignkeys.html?  It would avoid explicit checks in the code and also avoid serious corruption issues down the road.

icubam/db/sqlite.py
```diff
@@ -12,52 +14,65 @@ class SQLiteDB:
...
+      dept: str,
+      city: str,
+      lat: float,
+      long: float,
+      telephone: str = 'NULL',
   ):
     """Add or update an ICU."""
-
-    # If not then add:
+    # If not then add. We don't use SQLAlchemy doesn't yet support on conflict
```
dulacarnold:
```suggestion
    # If not then add. We don't use SQLAlchemy as it doesn't yet support on conflict
```

icubam/db/sqlite.py
```diff
@@ -143,11 +153,11 @@ def get_icus(self):
...
-         ON users.icu_id = icus.icu_id""",
-      self._conn,
+        """SELECT users.icu_id, icu_name, name, users.telephone, description
+           FROM users
+           JOIN icus
+           ON users.icu_id = icus.icu_id""",
+        self._conn,
     )
 
   def get_bedcount(self, icu_ids: Sequence = None, max_ts=None):
```
dulacarnold: IS this something we can do in SQLAlchemy?  This was the piece of code that originally made me want to stop writing queries by hand.

sgirgin:
> Overall this is great. We clearly have different code formatters, I'm using pyblack but happy to change, what are you using (linter/formatter)?

pyformat, https://pypi.org/project/pyformat/

icubam/db/sqlite.py
sgirgin: We already have one in users table, also added to bed_updates.

icubam/db/sqlite.py
sgirgin: Sure, converted it to a more sane form. Please see the note regarding the update_ts.

PR #46, opened 2020-03-27
draft DB wipe script

rprimet:
A draft script that wipes out confidential info from a DB dump.

Might have to wait till we move on to SQLAlchemy if we envision plugging another DB?


scripts/wipe_sensitive_info_db.py
```diff
@@ -0,0 +1,55 @@
...
+    conn.close()
+
+
+def main(argv):
+    if len(argv) < 2:
+        print(f"Usage: {argv[0]} DB_PATH")
+        sys.exit(-1)
+
+    if not FLAGS.quiet:
+        if not click.confirm("WARNING: THIS WILL WIPE THE DATABASE IN-PLACE. CONTINUE?", err=True):
```
VincentRouvreau: import click and added in requirements.txt just for the user to confirm the message ? I would have use stdin, but ok.

PR #49, opened 2020-03-28
backoffice

gqkc:
- Creation de la partie back office 
- Liste des users
- script to populate user

### DIFF
--- a/backoffice/model/encoder.py
+++ b/backoffice/model/encoder.py
@@ -1,0 +1,33 @@
+import json
+
+from sqlalchemy.ext.declarative import DeclarativeMeta
+
+
+class AlchemyEncoder(json.JSONEncoder):
+
+  def default(self, obj):
+    if isinstance(obj.__class__, DeclarativeMeta):
+      # an SQLAlchemy class
+      fields = {}
+      for field in [x for x in dir(obj) if not x.startswith('_') and x != 'metadata']:
+        data = obj.__getattribute__(field)
+        try:
+          json.dumps(data)  # this will fail on non-encodable values, like other classes
+          fields[field] = data
+        except TypeError:
+          fields[field] = None
+      # a json-encodable dict
+      return fields
+
+    return json.JSONEncoder.default(self, obj)
+
+from sqlalchemy.inspection import inspect
+
+class Serializer(object):
+
+    def serialize(self):
+        return {c: getattr(self, c) for c in inspect(self).attrs.keys()}
+
+    @staticmethod
+    def serialize_list(l):
+        return [obj.serialize() for obj in l]
--- a/backoffice/model/user.py
+++ b/backoffice/model/user.py
@@ -1,0 +1,15 @@
+from sqlalchemy import Column, Integer, String
+from sqlalchemy.ext.declarative import declarative_base
+
+from backoffice.model.encoder import Serializer
+
+Base = declarative_base()
+
+
+class User(Base, Serializer):
+  __tablename__ = 'user'
+  id = Column(Integer, primary_key=True)
+  name = Column('name', String(255))
+
+  def to_json(self):
+    return {c.name: getattr(self, c.name) for c in self.__table__.columns}
--- a/backoffice/www/handlers/base.py
+++ b/backoffice/www/handlers/base.py
@@ -1,0 +1,20 @@
+import tornado.locale
+import tornado.web
+
+
+class BaseBOHandler(tornado.web.RequestHandler):
+  """A base class for handlers."""
+
+  COOKIE = 'id'
+
+  def get_template_path(self):
+    return 'backoffice/www/templates/'
+
+
+  def get_current_user(self):
+    return self.get_secure_cookie("user")
+
+  def get_user_locale(self):
+    locale = self.get_query_argument('hl', default=None)
+    # We fallback to Accept-Language header.
+    return tornado.locale.get(locale) if locale else None
--- a/backoffice/www/handlers/home.py
+++ b/backoffice/www/handlers/home.py
@@ -1,0 +1,12 @@
+import tornado
+from tornado_sqlalchemy import SessionMixin
+
+from backoffice.www.handlers.base import BaseBOHandler
+
+
+class HomeBOHandler(SessionMixin, BaseBOHandler):
+  ROUTE = '/'
+
+  @tornado.web.authenticated
+  def get(self):
+    self.render("home.html")
--- a/backoffice/www/handlers/login.py
+++ b/backoffice/www/handlers/login.py
@@ -13,7 +13,7 @@ class LoginHandler(BaseBOHandler):
 
   def post(self):
     self.set_secure_cookie("user", self.get_argument("name"))
-    self.redirect("/")
+    self.redirect("/list_users")
 
 class AuthLogoutHandler(BaseBOHandler):
   def get(self):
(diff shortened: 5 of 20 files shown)


backoffice/www/handlers/base.py
```diff
@@ -0,0 +1,20 @@
...
+
+class BaseBOHandler(tornado.web.RequestHandler):
+  """A base class for handlers."""
+
+  COOKIE = 'id'
+
+  def get_template_path(self):
+    return 'backoffice/www/templates/'
+
+
```
fpquintao: it seems like we have more blank lines here than needed (?)
3 here, 1 between the two methods below.

backoffice/www/handlers/base.py
```diff
@@ -0,0 +1,20 @@
+import tornado.locale
+import tornado.web
+
+
+class BaseBOHandler(tornado.web.RequestHandler):
+  """A base class for handlers."""
+
+  COOKIE = 'id'
```
fpquintao: not used?

backoffice/www/handlers/home.py
```diff
@@ -0,0 +1,12 @@
...
+
+from backoffice.www.handlers.base import BaseBOHandler
+
+
+class HomeBOHandler(SessionMixin, BaseBOHandler):
+  ROUTE = '/'
+
+  @tornado.web.authenticated
+  def get(self):
+    self.render("home.html")
```
fpquintao: github complanining of "no newline at the end of the file".

backoffice/model/encoder.py
```diff
@@ -0,0 +1,33 @@
...
+from sqlalchemy.inspection import inspect
+
+class Serializer(object):
+
+    def serialize(self):
+        return {c: getattr(self, c) for c in inspect(self).attrs.keys()}
+
+    @staticmethod
+    def serialize_list(l):
+        return [obj.serialize() for obj in l]
```
fpquintao: github complaining of missing line at the end of of the file.

backoffice/model/user.py
```diff
@@ -0,0 +1,15 @@
...
+Base = declarative_base()
+
+
+class User(Base, Serializer):
+  __tablename__ = 'user'
+  id = Column(Integer, primary_key=True)
+  name = Column('name', String(255))
+
+  def to_json(self):
+    return {c.name: getattr(self, c.name) for c in self.__table__.columns}
```
fpquintao: missing new line.

backoffice/www/handlers/login.py
```diff
@@ -0,0 +1,21 @@
...
+  ROUTE= "/login"
+
+  def get(self):
+    self.write('<html><body><form action="/login" method="post">'
+               'Name: <input type="text" name="name">'
+               '<input type="submit" value="Sign in">'
+               '</form></body></html>')
+
+  def post(self):
+    self.set_secure_cookie("user", self.get_argument("name"))
```
fpquintao: so users can always login? We don't even check if the user exists?

Fine for me, it's more out of curiosity; if this is the case, we don't even need a login handler. I guess the server will be running under a service that is only reachable by a few people who have the IP whitelisted in Amazon. I think this is a reasonable way to start.

backoffice/www/handlers/user.py
```diff
@@ -0,0 +1,31 @@
...
+    self.write('{} users so far!'.format(count))
+
+
+class UserJson(SessionMixin, BaseBOHandler):
+  ROUTE = '/users_json'
+
+  @tornado.web.authenticated
+  def get(self):
+    users = self.session.query(User).all()
+    # self.write({"users": users},)
```
fpquintao: remove commented line?

backoffice/www/server.py
```diff
@@ -0,0 +1,58 @@
...
+    self.db = SQLAlchemy("sqlite+pysqlite:///{}".format(self.config.db.sqlite_path))
+    self.make_app()
+
+  def add_handler(self, handler, **kwargs):
+    route = os.path.join("/", handler.ROUTE)
+    self.routes.append((route, handler, kwargs))
+    logging.info("{} serving on {}".format(handler.__name__, route))
+
+  def make_app(self):
+    self.add_handler(LoginHandler)
```
fpquintao: super nice.

backoffice/www/server.py
```diff
@@ -0,0 +1,58 @@
...
+    self.add_handler(ListUserHandler)
+    self.add_handler(UserJson)
+    self.add_handler(CreateUserHandler)
+    self.add_handler(HomeBOHandler)
+
+  def run(self):
+    logging.info(
+      "Running {} on port {}".format(self.__class__.__name__, self.port)
+    )
+    engine = create_engine("sqlite+pysqlite:///{}".format(self.config.db.sqlite_path), encoding='utf-8', echo=True)
```
fpquintao: can we have in the constructor:

self.config ....
...
self.sqlite_path = "sqlite+pysqlite:///{}".format(self.config.db.sqlite_path)

then we can use it to initialise self.db (line 23) and the engine here.

backoffice/www/server.py
```diff
@@ -0,0 +1,58 @@
...
+      "login_url": "/login",
+      "static_path": "backoffice/www/static",
+      "db": self.db
+    }
+    tornado.locale.load_translations('icubam/www/translations')
+    app = tornado.web.Application(self.routes, **settings)
+    app.listen(self.port)
+
+    io_loop = tornado.ioloop.IOLoop.current()
+    # for callback_obj in self.callbacks:
```
fpquintao: should we drop these commented lines?

backoffice/www/static/list_users.js
```diff
@@ -0,0 +1,25 @@
...
+$(document).ready(function() {
+  $('#example').DataTable({
+    ajax: {
+      url: '/users_json',
+      data: function(d){
+            console.log(d)
+            return d
+        },
+    },
+    /*columnDefs: [ {
```
fpquintao: drop commented lines?

backoffice/www/handlers/user.py
```diff
@@ -0,0 +1,31 @@
...
+    self.write({"data": User.serialize_list(users)})
+
+
+class ListUserHandler(SessionMixin, BaseBOHandler):
+  ROUTE = '/list_users'
+
+  @tornado.web.authenticated
+  def get(self):
+    self.render("list_users.html")
```
fpquintao: I feel like this is missing the data? It's just invoking the template but we should pass the list of users that currently exist, like it's done in the /users_json handler (which I believe it's just for debug, which is great to have)

backoffice/www/handlers/home.py
```diff
@@ -0,0 +1,12 @@
+import tornado
```
fpquintao: I guess the backoffice should live inside the icubam folder.

backoffice/www/handlers/login.py
rprimet: Hmmm I'm out of the loop but that might warrant confirmation by @tehf0x and/or @olivierteboul?

scripts/populate_user.py
```diff
@@ -0,0 +1,31 @@
+from absl import app
+from absl import flags
+from sqlalchemy import create_engine
```
fpquintao: missing sqlachemy in the requirements.txt file?

gqkc:
Sorry, I resolved a conversation unintentionally 
I said we want to check if the user exists of course. This will be open to external users accepted by the admin.

gqkc:
I added a script to populate users in scripts

dulacarnold:
This has been cut up into other PRs.

PR #51, opened 2020-03-28
testing WWWServer

olivierteboul:


icubam/www/handlers/update.py
```diff
@@ -60,7 +60,7 @@ def get_icu_data_by_id(self, icu_id, def_val=0):
     input_data = self.token_encoder.decode(user_token)
 
     if input_data is None:
-      return self.redirect('/error')
+      return self.set_status(400)
```
VincentRouvreau: Wouldn't it be better to return 404 ? Be sure that returning 400 is not a way to tell hackers "try again".

icubam/www/handlers/update.py
```diff
@@ -72,7 +72,7 @@ def get_icu_data_by_id(self, icu_id, def_val=0):
 
     except Exception as e:
       logging.error(e)
-      return self.redirect('/error')
+      return self.set_status(400)
```
VincentRouvreau: Wouldn't it be better to return 404 ? Be sure that returning 400 is not a way to tell hackers "try again".

icubam/www/handlers/update.py
olivierteboul: good point ! doing it now

olivierteboul:
Thanks Vincent for the comments, very good point

PR #57, opened 2020-03-28
Backoffice init

gqkc:
home handler for back office


icubam/backoffice/server.py
```diff
@@ -0,0 +1,45 @@
+import os.path
+
+import tornado.ioloop
+import tornado.locale
+import tornado.web
+from absl import logging
+from tornado_sqlalchemy import SQLAlchemy
```
olivierteboul: this might indeed be interesting to have an async driver. but for now i would rather just use our DB abstraction, that now uses SQLAlchemy. So you can replace this line by `from tornado.db import sqlite`

icubam/backoffice/server.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class BackOfficeServer:
+  """Serves and manipulates the Backoffice ICUBAM."""
+
+  def __init__(self, config, port):
+    self.config = config
+    self.port = port
+    self.routes = []
+    self.db = SQLAlchemy("sqlite+pysqlite:///{}".format(self.config.db.sqlite_path))
```
olivierteboul: then merely use `sqlite.SQLDB(self.config.db.sqlite_path)`. This way someone can improve the db without having to touch this part of code.
But again, I fully agree, it is probably meaningful to look at the tornado_sqlalchemy package

PR #58, opened 2020-03-28
Added email field to users table.

sgirgin:
We are not using it yet.

### DIFF
--- a/icubam/db/sqlite.py
+++ b/icubam/db/sqlite.py
@@ -5,7 +5,7 @@ import time
 import pandas as pd
 from sqlalchemy import create_engine, desc, join
 from sqlalchemy import Table, Column, MetaData, ForeignKey, UniqueConstraint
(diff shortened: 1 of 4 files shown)


dulacarnold:
Hey Sertan, nice, since there is a DB migration step, do you think you could also add the following:
Column("n_ncovid_free", Integer), in bed_updates
And a boolean column 'active' that defaults to True to icus?

Normally I would prefer a seperate PR but because of the migration it might be easier to push all these changes at once.

sgirgin:
> n_ncovid_free

Sure, done.

dulacarnold:
I'm sorry, I got lost, we already have n_ncovid_free, we need to add n_ncovid_occ.  Sorry about that!  Otherwise LGTM.

sgirgin:
No worries, we probably don't need this change since it will be superseded
by the store.

PR #59, opened 2020-03-28
Add staging mode for sms scheduler

VincentRouvreau:


VincentRouvreau:
Workflow does not work for a fork...

PR #60, opened 2020-03-28
Simplified form pre-filling a bit.

dulacarnold:
@olivierteboul there is a bug with the time text generation however, let's sync tomorrow.


olivierteboul:
i sent you a fix already. thanks for catching it

PR #64, opened 2020-03-29
Add user list 

gqkc:
adding user list handler


icubam/backoffice/handlers/base.py
```diff
@@ -0,0 +1,17 @@
+import tornado.locale
```
fpquintao: this file already exists in the codebase. Can you sync your client?

icubam/backoffice/handlers/home.py
```diff
@@ -0,0 +1,10 @@
+from tornado_sqlalchemy import SessionMixin
```
fpquintao: same here; file already exists.

icubam/backoffice/server.py
```diff
@@ -0,0 +1,48 @@
+import os.path
```
fpquintao: same here

dulacarnold:
@fpquintao should we close or keep this open?

fpquintao:
Can we keep it open for now?
There are some files here that will be useful soon.

fpquintao:
I will close this branch since we have one "list users" in place in the backoffice right now.

PR #77, opened 2020-03-29
Login/Logout and class hierarchy for backoffice handlers; using list_users as an example for the idea

fpquintao:


fpquintao:
@olivierteboul can you take a look?
I'm switching gears now to corp work.
There's a TODO for you in the code.

Thanks,
---q

icubam/backoffice/handlers/base.py
```diff
@@ -10,6 +11,8 @@ class BaseHandler(tornado.web.RequestHandler):
   def initialize(self, config, db):
     self.config = config
     self.db = db
+    # TODO(olivier): this should come from the server.
```
olivierteboul: ack.

icubam/backoffice/handlers/base.py
```diff
@@ -21,3 +24,15 @@ def get_user_locale(self):
     locale = self.get_query_argument('hl', default=self.config.locale)
     # We fallback to Accept-Language header.
     return tornado.locale.get(locale) if locale else None
+
+  def set_default_headers(self):
```
olivierteboul: nice, btw, after we should try to figure out how to have it glued to icubam website, still running another process..

icubam/backoffice/handlers/store.py
```diff
@@ -0,0 +1,43 @@
...
+
+  """Invokes the GET operation on the context of the user."""
+  def getForUser(self, user):
+    pass
+
+  """Invokes the POST operation on the context of the user."""
+  def postForUser(self, user):
+    pass
+
+  def validate_and_call(self, callback):
```
olivierteboul: i believe the `@tornado.web.authenticated` already checks if there is a current user, via cookie.
Then one way is to overload get_current_user() and add the user in like self._user ? Then you can simple add the decorator whereever you want authentication + access to the user...

icubam/backoffice/handlers/base.py
fpquintao: :)

icubam/backoffice/handlers/base.py
fpquintao: do you want a single server?
Fine for me, I will have to check the structure of the website. Let's open issue/ or something to track that.

icubam/backoffice/handlers/store.py
fpquintao: that's great. I didn't realize we can override get_current_user to return whatever we want :)
(too much C++ in the last days).

PR #79, opened 2020-03-29
replace image with external link

tgy:


tgy:
@VincentRouvreau i did `rm -rf images` so i expect that directory will disappear

VincentRouvreau:
> @VincentRouvreau i did `rm -rf images` so i expect that directory will disappear

Sounds good to me.

PR #82, opened 2020-03-30
Improve packaging: use setup.cfg

Deathn0t:


icubam/__init__.py
```diff
@@ -0,0 +1,14 @@
+import pkg_resources
+from os import path
+from setuptools.config import read_configuration
+
+def _extract_version(package_name):
```
rth: Ideally __init__ should not import setuptools or load any files. 

The standard way is rather to indicate `__version__` here and load it from setup.py.

rth:
(Also please remove the ` .DS_Store` file).

icubam/__init__.py
Deathn0t: This is what I usually do, but I thought it was better to put it in the `setup.cfg`?

rth: You can indeed put it in `setup.cfg`, but the version in `__init__.py` should be self contained and not depend on any other file. You are not guaranteed that `setup.cfg` will be there once you pip install the package. Also performance wise it's bad. Which could mean two files that need editing on version update.

https://github.com/pypa/setuptools_scm might be useful to automate versioning somewhat BTW.

setup.py
```diff
@@ -10,4 +10,4 @@
         # If any package contains *.txt or *.rst files, include them:
         "": ["*.env"]
```
rth, line 11: Please remove from `setup.py` information that is currently in `setup.cfg` such as version, description, license, name.

icubam/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+    try:
+        return pkg_resources.get_distribution(package_name).version
+    except pkg_resources.DistributionNotFound:
+        _conf = read_configuration(path.join(
+            path.dirname(path.dirname(__file__)), 'setup.cfg')
+        )
+        return _conf['metadata']['version']
+
+__version__ = _extract_version('icubam')
```
rth, line 14: To move forward on this let's just put,
```suggestion
__version__ = '0.1.2'
```
(same as in resources/config.toml) for now and address this versioning issue in a separate PR closing https://github.com/icubam/icubam/issues/121

setup.cfg
```diff
@@ -0,0 +1,39 @@
+[metadata]
+name = icubam
+description = "ICU Bed Activity Monitor."
+version = 0.1.0
```
rth, line 4:
```suggestion
version = 0.1.2
```
(same as in resources/config.toml)

setup.cfg
```diff
@@ -0,0 +1,39 @@
...
+    Natural Language :: English
+    License :: OSI Approved :: Apache Software License
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.8
+    Intended Audience :: Healthcare Industry
+
+
+[options]
+install_requires =
```
rth, line 20: Either we keep it here, and remove `requirements.txt` or maybe there is an option to load `requirements.txt` here (or in setup.py).  I still think having a separate `requirements.txt` file is useful.

In either case, I would not keep both, as they are redundant and annoying to maintain in sync.

If we want to merge this quickly maybe let's remove it for now, and think about it in a follow up PR?

rprimet: @rth what I've seen done is the opposite way round i.e. keep dependencies in `setup.cfg` and replace the whole contents of `requirements.txt` by `-e .`. That way `pip install -r requirements.txt` and `pip install -e .` are equivalent and pull deps as well. How does that sound? Anyway I  agree that it's mostly nitpicking and can be done after this PR has been merged?

icubam/www/server.py
```diff
@@ -1,59 +0,0 @@
-import os.path
```
rth, line 1: Removing this file looks like an accident?

setup.cfg
rth: That could also be fine, as long as we don't keep maintaining both. Though I'm not familiar with all deployment constraints.

rth:
OK, I pushed a few commits to illustrate what this could look like if we moved all config from `setup.py` to `setup.cfg` and also included `requirements.txt` there.

There is some improvement with respect to the current approach, but I also can't say that there is currently a problem that really needs fixing. There is a small risk of install issues if I messed up somewhere in `setup.cfg`.

This is ready for review now. Looking for other opinions.

dulacarnold:
I'm going to close this as it's become stale and no one is available to sync this.

PR #92, opened 2020-03-30
Migrate the message and www server to new store

olivierteboul:


icubam/db/store.py
```diff
@@ -229,6 +240,13 @@ def get_managed_icus(self, manager_user_id: int) -> Iterable[ICU]:
 
   # User related methods.
 
+  def add_default_admin(self) -> int:
+    """Creates a default 'admin/admin' user."""
+    name = 'admin'
+    hash = self.get_password_hash(name)
```
sgirgin: Can we possibly pick a less predictable password, just in case?

icubam/www/handlers/home.py
```diff
@@ -22,45 +24,71 @@ class HomeHandler(base.BaseHandler):
...
-        cluster_id[icuid] = city
-    return coords, cluster_id
-
-  def get_phones(self):
+
+  def groupby(self):
     result = collections.defaultdict(list)
-    for index, row in self.icus_df.iterrows():
-      result[row['icu_id']].append(row['telephone'])
+    for icu in self.icus.values():
```
sgirgin: With the function that coverts DB objects to dicts, I'm curious if this may also work for grouping?

https://stackoverflow.com/questions/20638006/convert-list-of-dictionaries-to-a-pandas-dataframe

icubam/db/store.py
```diff
@@ -496,3 +502,20 @@ def create_store_for_sqlite_db(cfg) -> Store:
...
+def get_column_names(obj):
+  """Returns the columns of a given class."""
+  return list(obj.__mapper__.columns.keys())
+
+def to_dict(obj):
+  """Turns a Base instance into a dictionary."""
+  columns = get_column_names(obj)
+  result = {}
+  for col in columns:
+    result[col] = getattr(obj, col)
```
fpquintao: cool!

icubam/www/handlers/home.py
olivierteboul: that is a good point. the function is not recursive for now. let's keep it for later though, since there are already a lot in the PR

fpquintao:
Merged on behalf of Olivier who had to step out a bit :)

PR #100, opened 2020-03-31
ability to active/deactive sms via handler

olivierteboul:
- along with test on the scduler
- the hander is not connected yet. PR to come
- wait for migration PR to get in before reviewing


icubam/messaging/handlers/onoff_test.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+from icubam.messaging.handlers import onoff
+
+class OnOffRequestTestCase(unittest.TestCase):
```
rth: Since tests are run with pytest, you don't need importing `unittest`:

```suggestion
class OnOffRequestTestCase:
```

icubam/messaging/handlers/onoff_test.py
```diff
@@ -0,0 +1,25 @@
...
+  def test_silence_request(self):
+    userid = 123
+    icuid = 773
+    on = False
+    delay = 34
+
+    request = onoff.OnOffRequest(
+      user_id=userid, icu_id=icuid, on=on, delay=delay)
+    encoded = request.to_json()
+    self.assertIsInstance(encoded, str)
```
rth: with pytest

```suggestion
    assert isinstance(encoded, str)
```

icubam/messaging/handlers/onoff_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+    decoded = onoff.OnOffRequest()
+    self.assertEqual(decoded.user_id, None)
+
+    decoded.from_json(encoded)
+    self.assertEqual(decoded.user_id, userid)
+    self.assertEqual(decoded.icu_id, icuid)
+    self.assertEqual(decoded.on, on)
+    self.assertEqual(decoded.on, on)
+    self.assertEqual(decoded.delay, delay)
```
rth:
```suggestion
    assert decoded.user_id == userid
    assert decoded.icu_id == icuid
    assert decoded.on == on
    assert decoded.on == on
    assert decoded.delay == delay
```

pytest will produce better error message than unittest anyway

icubam/messaging/scheduler_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+class MockQueue:
+  def __init__(self):
+    self.data = []
+
+  def put(self, data):
+    self.data.append(data)
+
+
+class SchedulerTestCase(absltest.TestCase):
```
rth: Same comments as above about `TestCase` and not using `self.assertEqual`

icubam/messaging/handlers/onoff.py
```diff
@@ -0,0 +1,50 @@
...
+
+  def to_json(self):
+    return json.dumps(dataclasses.asdict(self))
+
+  def from_json(self, encoded):
+    self.__init__(**json.loads(encoded))
+
+
+class OnOffHandler(tornado.web.RequestHandler):
+
```
sgirgin: Can you add a comment about the purpose of the handler? It will help if one doesn't know the context,

icubam/messaging/scheduler.py
```diff
@@ -1,74 +1,103 @@
...
+      self.unschedule(msg.user_id, msg.icu_id)
+
+    handle = io_loop.call_later(delay, self.may_send, msg)
+    self.timeouts[key] = ScheduledMessage(handle, msg, when)
+    logging.info('Scheduling {} in {}s.'.format(msg.icu_name, delay))
+    return True
+
+  def schedule(self, user, icu_id: int, delay: Optional[int] = None) -> bool:
+    user_icus = {i.icu_id: i for i in user.icus}
+    if not user.is_active or icu_id not in user_icus:
```
sgirgin: Do we want to check if ICU is active?

olivierteboul:
thanks @rth pytest is used in the deployment process, however an external user forking the project my not do it, so i prefer keeping the most generic form. It is also more explicit to me.

icubam/messaging/scheduler.py
```diff
@@ -1,74 +1,103 @@
...
+      user_id = user.user_id
+      logging.info(f'Cannot send message to user {user_id} in icu {icu_id}')
+      return False
+
+    url = self.updater.get_user_url(user, icu_id)
+    msg = message.Message(icu_id, user, url)
```
sgirgin: Since you are changing Message anyway, it may be cleaner if we pass the icu instead of icu_id here. That will also simplify the checks there.

rth:
> pytest is used in the deployment process, however an external user forking the project my not do it, so i prefer keeping the most generic form. It is also more explicit to me.

It's your decision, but if a user managed to install 15 dependencies from `requirements.txt`, asking to install one more for testing is not really going to matter. It's fairly standard for python packages, I don't think it will be a blocker for contributors. 

Pytest does produce much better error messages when tests fails and is less verbose for maintainers. Also stuff like parametric tests, and other pytest features do help with increasing the test coverage.

olivierteboul:
thanks for the suggestion sertan. i have improved the code with `icu` instead of `icu_id` + added more tests

PR #112, opened 2020-03-31
fix in a debug function + test to catch it

olivierteboul:


icubam/messaging/scheduler.py
```diff
@@ -57,7 +57,8 @@ def schedule_message(
 
     when = delay + time.time()
     timeout = self.timeouts.get(msg.key, None)
-    if timeout is not None and when > timeout.when:
+    margin = 10
```
fpquintao: perhaps a little comment about the margin idea.

PR #113, opened 2020-03-31
ENH Add /version endpoint

rth:
This adds the `/sanity-checks` endpoint. 

The current output is a JSON (that should be pretty-printed by the browser) with,
```json
{
    "data": {
        "package": {"version": "0.1", "git-hash": "5decb5e"},
        "db": {
            "icus.count": 11,
            "bed_counts.count": 5,
            "bed_counts.n_covid_occ": 52,
            "bed_counts.n_covid_free": 49,
            "users.count": 22,
            "users.is_admin.count": 6,
            "regions.count": 1,
        },
        "server": {"load-average": [0.0, 0.0, 0.0]},
    }
}
```
please let me know if it would be helpful to add other things there (or if some of this information is not relevant).

**Edit:** it is pretty-printed in Firefox, not Chrome. I can render it as an HTML page it it's easier, but I would like to get some feedback first. Can't really write useful tests in that case.

Also I'm not sure if the general goal is to,
 - run those DB queries and check that the numbers are consistent
 - or check that none of them raises exceptions, in which case maybe we should wrap each call in some form of `try: ... except:..`

The git commit hash, is obtained by running git (assumes install in develop mode), otherwise we would need to somehow include it in the package information (e.g. using tools like versioneer or setuptools_scm, I haven't looked in details).


.github/workflows/pythontest-linux.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         # Emit pickled gsheets testing credentials from CI secret
         python -c "import pickle; import base64; import os; tok=pickle.loads(base64.b64decode(os.getenv('GSHEETS_TOKEN'))); tfile = open('token.pickle', 'wb'); pickle.dump(tok, tfile)"
         pip install pytest
-        pytest .
+        pytest --doctest-modules icubam/
```
rth: Ideally this option should go in `setup.cfg` when it's added.

dulacarnold: I think we're not set up with a good versioning system.  We started putting versions in the config.toml but that was a bad idea, I think bumping the setup.py version every time we deploy could work.  What would you suggest?

dulacarnold: Ok, sorry I hadn't read through the json.  I think it's better to not print any information about beds etc. but to just run those calls inside try/catch blocks to make sure nothing bad is happening.  Ideally we'd have the git hash and package version.  Currently the version is accessible from the config with self.config.version, but I think that ought to be moved to icubam.__version__ or something easier to maintain.

olivierteboul: Thanks a lot Roman for the PR. I agree wih Gabriel (maybe we did not communicate clearly at first, sorry about that). but just having git commit hash, maybe commit tag and version would be perfect. Which simplifies a lot the PR. Then the route could be something like version/ ?

rth: Thanks for the feedback! Sure will make the changes -- that will simplify things indeed.

rth:
Thanks I have updated this PR following your comments. For now this takes the version from `icubam.__init__` and I agree that versioning can be improved, but let's do that in a follow up PR. I opened https://github.com/icubam/icubam/issues/121 about it.

The current JSON output is,
```
{"data": {"package": {"version": "0.1.2", "git-hash": "d85729e"}, "errors": {}}}
```
and `errors` will contain error messages if one of those `get_*` functions in `Storage` fails.

.github/workflows/pythontest-linux.yml
olivierteboul: hi roman, it s nice, but i am not sure this route should be only accessible to authenticated people. That is why i suggest to simplify it even more with no db access. Sorry to make you erase what you wrote. But please double check this statement with Romain before.
Thanks

olivierteboul: i  mean errors is not a bad idea at all... but double check how this would be accessed first

rth:
> hi roman, it s nice, but i am not sure this route should be only accessible to authenticated people. That is why i suggest to simplify it even more with no db access. Sorry to make you erase what you wrote. But please double check this statement with Romain before.
Thanks
> i mean errors is not a bad idea at all... but double check how this would be accessed first

So the question is who is going to use it and with what motivation. And what should happen if there is an issue in DB query if we remove the authentication via the cookie,
 - this endpoint just fails and you get the traceback in the log
 - this endpoint `try: except:` all  DB queries, but then the questions is  what happens if there is an error. There is no log so it's kind of hard to debug the issue.
 - this endpoint just returns the version with no DB queries. Which is also fine, but that's contrary to what the initial Task description says.

Generally it would be good to use Github issue to more clearly document what we want before implementing it.

rth:
> But please double check this statement with Romain before.

@rprimet What do you think (cf discussion above). I'm still not sure I fully understand the use case of this endpoint.

rprimet:
Hi @rth I think this endpoint's goal is to provide a basic health check and version info, but without leaking too much context (the idea being that deployers can monitor it with a regular "ping" and ensure it returns 200, and give a quick way of ensuring which version of the code runs on a given deployment). 

I agree that it should be unauthenticated since that would be simpler to monitor (and hence it should not leak information beyond "is this running somewhat as expected" and "what version of the code is this deployment running"?)

rth:
> (the idea being that deployers can monitor it with a regular "ping" and ensure it returns 200, and give a quick way of ensuring which version of the code runs on a given deployment).

@rprimet So, if it runs those DB queries and returns a 500 Internal Server Error if something fails (with the traceback in the server log), without leaking any information in the returned JSON would that be OK?  I should I not run those DB queries at all (in which case it wouldn't really check anything)?

rth:
Or should there be 2 endpoints,
 - /version  - which always succeeds
 - /sanity-checks - which runs those DB queries and fails if something went wrong
?

rprimet:
I'd go with a simple /ping endpoint (or leave the current name) that returns the version (commit hash if installed from source if such a thing is possible) and either runs the query and returns a 500 in case of error, or runs a very simple query and add e.g. the last bed_update timestamp. The only thing I want to stress is to avoid long-running queries, otherwise your solution (error 500 + traceback in logs) sounds great.

rth:
OK, got it. Thanks!

rth:
OK I pushed a fix to address comments, please have a look. Current output (unauthenticated),
```bash
$ http GET http://localhost:8888/version
HTTP/1.1 200 OK
Content-Length: 104
Content-Type: application/json; charset=UTF-8
Date: Wed, 01 Apr 2020 11:25:54 GMT
Etag: "fda3dd818a0854d9babedfde426e1cbae08a0d66"
Server: TornadoServer/6.0.3

{
    "data": {
        "bed_counts.last_modified": "2020-03-31 16:35:35",
        "git-hash": "2c18a80",
        "version": "0.1.2"
    }
}
```

> runs a very simple query and add e.g. the last bed_update timestamp. The only thing I want to stress is to avoid long-running queries

Currently it only runs, `get_bed_counts` -- shouldn't take too long I think.

dulacarnold:
Agree with all above, it should run the standard query for the map: self.db.get_visible_bed_counts_for_user(None, force=True) in a try catch and return in the json whether that succeeded or not (check len > 0 or Exception).  SGTY?

setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 setup(name='icubam',
-      version='0.1',
+      version='0.1.2',
```
dulacarnold, line 4: I'll make sure to bump this with the deploy version.

rprimet:
Seems good! I don't know about the perf characteristics of the query so if @olivierteboul is fine with that I'd say great.

Something else to watch for: could we ensure somehow that this route is uncached (I'm not that familiar with Tornado so I didn't know where to look in the app). Anyway, setting a header  like `Cache-Control' = 'no-cache, no-store, must-revalidate'` in the reply might help?

icubam/www/handlers/version.py
```diff
@@ -0,0 +1,36 @@
...
+    """Get sanity checks data"""
+    data = {}
+    data['version'] = icubam.__version__
+    try:
+      git_hash = check_output(['git', 'rev-parse', '--short', 'HEAD'],
+                              encoding='utf-8').strip()
+    except Exception:
+      git_hash = 'NA'
+    data['git-hash'] = git_hash
+    bed_counts = self.db.get_bed_counts()
```
dulacarnold, line 30: Can we wrap this in a try/catch and specify in the returned data whether it returned correctly && len(bedcounts) >0, or incorrectly (len == 0 or Exception)?

setup.py
dulacarnold: (to be clear, nothing to do here).

rth:
> Agree with all above, it should run the standard query for the map: self.db.get_visible_bed_counts_for_user(None, force=True) in a try catch and return in the json whether that succeeded or not (check len > 0 or Exception). SGTY?

That's similar to what was done before, before I reverted it following comments above. If we want to have a traceback in the log and a 500 error status on failed query, the easiest is to run the query without the `try: except:`..

icubam/www/handlers/version.py
rth: We then would need specific code to log the traceback + error message, and manually return HTTP 500 error status. See https://github.com/icubam/icubam/pull/113#issuecomment-607199272

rprimet: I would say let's do the simplest thing that can work (i.e. do not catch the exception)? We don't want to supply too much context anyway... My 2cents :)

dulacarnold:
OK pour me. LGTM.

icubam/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.1.2'
```
dulacarnold, line 1: Maybe this should be kept in the setup.py, and it will get pushed on pip install.

rth: Yes, in the end it should be. I opened https://github.com/icubam/icubam/issues/121 for it.

I'll open a follow up PR to address it.

rth:
A added a header to avoid caching for sure. Good to merge? I'll open a follow up PR reduce the number of places version has to be specified manually.

rprimet:
LGTM!

PR #117, opened 2020-03-31
Update dev install procedure

VincentRouvreau:
* Update dev install procedure
* Use `click` to accept to populate_db
* Use flags for `.env` file in populate_db_fake
* A new script to get id url - cf. http://localhost:8888/update?id=<A_VERY_LONG_ID>

### DIFF
--- a/dev_install.md
+++ b/dev_install.md
@@ -15,7 +15,7 @@ Steps:
 
 ### Configuration
 
-Create a `.env` file at root of the project containing the following keys:
+Create a `resources/.env` file containing the following keys:
 ```
 SHEET_ID=
 TOKEN_LOC=
@@ -28,19 +28,23 @@ NX_KEY=
 NX_API= 
 TW_KEY=
 TW_API=
+DB_SALT=
 ```
+
+N.B.: You can name and move this file as you want but you will have to add
+`--dotenv_path=path/to/my_icubam.env` to the scripts when launching them.
 
 ### Pre-populate DB with test data
 
 Create a fake database in order to be able to play with it:
-`python scripts/populate_db_fake.py`
+`python scripts/populate_db_fake.py --config=resources/config.toml --mode=dev`
 
-The databse will be named `test.db`.
+The database will be named `test.db`, cf. `resources/config.toml`.
 
 ## Running locally
 
 Start the server locally:
-`python scripts/run_server.py`
+`python scripts/run_server.py --config=resources/config.toml --mode=dev`
 
 Will produce the following logs:
 ```
@@ -54,6 +58,6 @@ Follow the proposed link `http://localhost:8888/update?id=<A_VERY_LONG_ID>`
 
 ## Running unit tests
 
-The unit tests require `TOKEN_LOC` to be set with a valid `token.pickle` file in the `.env` file.
+The unit tests require `TOKEN_LOC` to be set with a valid `token.pickle` file in the `resources/.env` file.
 
 To start the tests, install `pytest` and run `pytest`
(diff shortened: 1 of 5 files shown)


dev_install.md
```diff
@@ -15,7 +15,7 @@ Steps:
 
 ### Configuration
 
-Create a `.env` file at root of the project containing the following keys:
+Create a `resources/.env` file containing the following keys:
```
dulacarnold: Since we're always passing this file as a flag anyways, can we just call it icubam.env?  Or is this a default path/

dev_install.md
```diff
@@ -28,19 +28,23 @@ NX_KEY=
 NX_API=
```
dulacarnold: We can get rid of NX_API as well.

dev_install.md
VincentRouvreau: This is the default value in some scripts (populate_db_fake, migrate_db, populate_db, synchronize_db), but  not for run_server. Argh...
Do you want me to set default value to `resources/icubam.env` for every body ? (+ run_server, schedule_sms, send_sms)

VincentRouvreau:
@dulacarnold Sorry, I was not finished with my PR when you reviewed it but I also wrote an utility to get url identifiers

scripts/get_id_url.py
```diff
@@ -0,0 +1,27 @@
...
+flags.DEFINE_string('dotenv_path', None, 'Optionally specifies the .env path.')
+flags.DEFINE_enum('mode', 'dev', ['prod', 'dev'], 'Run mode.')
+FLAGS = flags.FLAGS
+
+def main(argv):
+  cfg = config.Config(FLAGS.config, mode=FLAGS.mode, env_path=FLAGS.dotenv_path)
+  sqldb = sqlite.SQLiteDB(cfg.db.sqlite_path)
+  icus = sqldb.get_icus()
+  encoder = token.TokenEncoder(cfg)
+  for _, icu_row in icus.iterrows():
```
olivierteboul: check updater.py, that have function that generates the urls.
And we migrated to the new store:
```
for user in self.db.get_users():
  for icu in user.icu:
   ...
```

VincentRouvreau: Maybe I need to wait for the new db format ?
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: users.icu_id
[SQL: SELECT users.icu_id, icu_name, name, users.telephone, description
           FROM users
           JOIN icus
           ON users.icu_id = icus.icu_id]
```

scripts/populate_db.py
```diff
@@ -1,6 +1,7 @@
 """This will populate the database from the google sheet."""
 from absl import app
 from absl import flags
+import click
```
olivierteboul, line 4: we would need to add this to the requirements.txt too.

however i am not even sure this script is used any more with the new db

VincentRouvreau: It is already installed because used by `scripts/wipe_sensitive_info_db.py`. Tell me if you want to delete this script.

scripts/get_id_url.py
VincentRouvreau: It works, thanks to the new `icubam.db.store`

dev_install.md
dulacarnold: If you feel like it's not too much probably good to standardize, and therefore might as well make the file visible.

scripts/populate_db.py
dulacarnold: I think Romain is updating it, so let's leave it for now.

dulacarnold:
@rprimet can you propose a PR for that? I'd like to merge the updated docs at least.

rprimet:
@dulacarnold OK, I'll submit another PR once this one is merged to separate both jobs.

Issue #108, opened 2020-03-31
test_gsheets.SQLiteDBTest test failure 


rth:
When following the `dev_install` a couple of tests fail,
```
$ pytest .
[...]
_______________________________________________________ SQLiteDBTest.test_icus ________________________________________________________

self = <icubam.db.test_gsheets.SQLiteDBTest testMethod=test_icus>

    def test_icus(self):
>     shdb = gsheets.SheetsDB(self.config.TOKEN_LOC, self.config.SHEET_ID)

icubam/db/test_gsheets.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <icubam.db.gsheets.SheetsDB object at 0x7f01ce21e940>, token_loc = '', sheet_id = ''

    def __init__(self, token_loc, sheet_id):
      """Given a token file and a sheet id, loads the sheet to be queried."""
      self._token_loc = token_loc
      self._sheet_id = sheet_id
    
>     with open(self._token_loc, "rb") as token:
E     FileNotFoundError: [Errno 2] No such file or directory: ''

icubam/db/gsheets.py:18: FileNotFoundError
_______________________________________________________ SQLiteDBTest.test_users _______________________________________________________

self = <icubam.db.test_gsheets.SQLiteDBTest testMethod=test_users>

    def test_users(self):
>     shdb = gsheets.SheetsDB(self.config.TOKEN_LOC, self.config.SHEET_ID)

icubam/db/test_gsheets.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <icubam.db.gsheets.SheetsDB object at 0x7f01ce03b0d0>, token_loc = '', sheet_id = ''

    def __init__(self, token_loc, sheet_id):
      """Given a token file and a sheet id, loads the sheet to be queried."""
      self._token_loc = token_loc
      self._sheet_id = sheet_id
    
>     with open(self._token_loc, "rb") as token:
[...]

PR #118, opened 2020-03-31
unit tests should pass on a vanilla clone

rprimet:
Unit tests should pass from a "vanilla" clone without the need for tweaking / setting a configuration, and should not require secrets. Hence, this PR sets a dummy `JWT_TOKEN` env var for testing, disables `test_gsheets` as it requires a secret and enables github actions on pull requests to ease collaboration.

### DIFF
--- a/.github/workflows/pythontest-linux.yml
+++ b/.github/workflows/pythontest-linux.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   test:
@@ -27,15 +27,10 @@ jobs:
     - name: Test with pytest
       shell: bash -l {0}
       env:
-        GSHEETS_TOKEN: ${{ secrets.gsheets_token }}
-        TOKEN_LOC: token.pickle
-        SHEET_ID: ${{ secrets.SHEET_ID }}
-        JWT_SECRET: ${{ secrets.JWT_SECRET }}
+        JWT_SECRET: faketest
       run: |
-        # Emit pickled gsheets testing credentials from CI secret
-        python -c "import pickle; import base64; import os; tok=pickle.loads(base64.b64decode(os.getenv('GSHEETS_TOKEN'))); tfile = open('token.pickle', 'wb'); pickle.dump(tok, tfile)"
         pip install pytest
-        pytest .
+        pytest --ignore icubam/db/test_gsheets.py
         pip install -e .
         python scripts/populate_db_fake.py --mode=dev
         timeout 5 python scripts/run_server.py || echo "Normal timeout" | grep "Normal timeout"
--- a/.github/workflows/pythontest-linux.yml
+++ b/.github/workflows/pythontest-linux.yml
@@ -30,7 +30,7 @@ jobs:
         JWT_SECRET: faketest
       run: |
         pip install pytest
-        pytest --ignore icubam/db/test_gsheets.py
+        pytest
         pip install -e .
         python scripts/populate_db_fake.py --mode=dev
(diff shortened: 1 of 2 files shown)


rprimet:
That would enable contributors to work on forks and have CI run but is still not ideal as a developer that checks out a local copy would still have to set `JWT_TOKEN` to run the tests on his/her machine (as per the instructions in `dev_install.md`, but still, no setup would be even better).

olivierteboul:
i agree with the gsheet_test.py. However, i would rather mock it than delete it. But we can remove it for now and put it back later.

.github/workflows/pythontest-linux.yml
```diff
@@ -27,15 +27,10 @@ jobs:
...
-        TOKEN_LOC: token.pickle
-        SHEET_ID: ${{ secrets.SHEET_ID }}
-        JWT_SECRET: ${{ secrets.JWT_SECRET }}
+        JWT_SECRET: faketest
       run: |
-        # Emit pickled gsheets testing credentials from CI secret
-        python -c "import pickle; import base64; import os; tok=pickle.loads(base64.b64decode(os.getenv('GSHEETS_TOKEN'))); tfile = open('token.pickle', 'wb'); pickle.dump(tok, tfile)"
         pip install pytest
-        pytest .
+        pytest --ignore icubam/db/test_gsheets.py
```
rth, line 33: Instead if ignoring them via CLI, maybe to add something along the lines of,

```py
ci_tokens = all([os.environ.get(key, False)
                 for key in ['GSHEETS_TOKEN', 'SHEET_ID', 'TOKEN_LOC]])

@unittest.skipIf(not ci-tokens,
                 "GSHEETS_TOKEN env variable not set")
```
https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures

to the two tests that fail, as otherwise users who clone the repo, wouldn't know to exclude this tests.

This closes https://github.com/icubam/icubam/issues/108 BTW.

rprimet: Thanks for the comments! 

The code would go into the gsheets unit test itself, right? So maybe rename `ci_tokens` to `gsheets_tokens`?

Usually devs who have the code set up on their machine and run the gsheets tests have a `token.pickle` file pointed to by `TOKEN_LOC` (it is only the current CI pipeline that stuffs it in a secret, base64-encoded, to avoid putting secrets in the tree). So how about restricting the tests to TOKEN_LOC and SHEET_ID? That would let forks / users with a blank config avoid running the tests while still failing loudly when it is expected to run (and it would keep running in CI)?

rth: > The code would go into the gsheets unit test itself, right? So maybe rename ci_tokens to gsheets_tokens?

Yes, there. Up to you, just to avoid confusion with `GSHEETS_TOKEN` env variable.

> So how about restricting the tests to TOKEN_LOC and SHEET_ID? That would let forks / users with a blank config avoid running the tests while still failing loudly when it is expected to run (and it would keep running in CI)?

Sounds good.

icubam/db/test_gsheets.py
```diff
@@ -1,18 +1,26 @@
...
 from icubam.db import gsheets
 from absl.testing import absltest
+import os
+import unittest
 
 
-class SQLiteDBTest(absltest.TestCase):
+class GsheetsTest(absltest.TestCase):
+  gsheets_tokens_def = all([os.environ.get(key, False)
+                           for key in ['SHEET_ID', 'TOKEN_LOC']])
```
rth, line 10: I would have though that this should be defined one line before the class, otherwise it would have to be accessed as `self.gsheets_tokens_def` which is not possible from a decorator.

Although this seems to pass tests, and the two tests are indeed skipped, so maybe it's fine.

PR #119, opened 2020-04-01
better ENV management in dev/prod

olivierteboul:


icubam/config.py
```diff
@@ -46,7 +58,10 @@ def _preprocess(self, conf):
...
     return result
 
-  def __getattr__(self, key: str):
+  def __getitem__(self, key: str):
     if key.upper() == key:
       return self.env.get(key)
     return self.conf.get(key)
+
+  def __getattr__(self, key):
+    return self[key]
```
rth, line 67: IMO the shortcuts,
 - cfg.key -> config value
 - cfg['key'] -> config value
 - cfg['KEY']  -> env variable 

can lead to confusion, I would have removed the last option in favor of `cfg.env.get('key')` explicitly. Env variables shouldn't be accessed that often anyway?

But that's a comment about existing code, not necessarily this PR

olivierteboul: it s a good point Roman. I would probably require a proper PR though

olivierteboul:
closing for now

PR #124, opened 2020-04-01
Backoffice: styles in form and tables

olivierteboul:


rth:
If I wanted to check this out, what is the URL endpoint for backoffice? It is started with the same `scripts/run_server.py` script, right?

olivierteboul:
python3 -m scripts.run_server --server=backoffice

Issue #121, opened 2020-04-01
Decide on a versioning system


rth:
As mentioned in https://github.com/icubam/icubam/pull/113#discussion_r401215066 by @dulacarnold 
> I think we're not set up with a good versioning system. We started putting versions in the config.toml but that was a bad idea, I think bumping the setup.py version every time we deploy could work (a).

Opening an issue to centralize the discussion around it, as there have been lots of comments in different places.

Earlier, @rprimet suggested to use https://github.com/warner/python-versioneer for it (b). Another alternative could be https://github.com/pypa/setuptools_scm (comparison with versioneer in https://github.com/icubam/icubam/pull/82) (c).

Also in https://github.com/icubam/icubam/pull/82 @Deathn0t proposed another solution to load version from `setup.cfg` (d).

So there are 4 possible approaches so far a-b-c-d.

rth:
Personally I think a) is fine as long as deployments are not too frequent. I'll investigate b and c to see if they could work in this use-case.

rth:
Version is currently set in 3 place,
 - setup.py (or setup.cfg)
 - `icubam/__init__.py` (after https://github.com/icubam/icubam/pull/113 is merged)
 - `resources/config.toml`

and ideally we would need to reduce this to 1 even for approach a).

Deathn0t:
Personally to manage projects simply I usually set manually the version in a `icubam/__version__.py` which is loaded in the `icubam/__init__.py`. And the `setup.py` import this `__version__` attribute from the icubam package. Till now I never used the `setup.cfg` just thought it was a more standard way so created one in the #82. I also usually put all requirements in a list inside the `setup.py` so that everything is in the same place. Then I create the release tag using [git-flow](https://danielkummer.github.io/git-flow-cheatsheet/), which is very easy and well organized for features, hot-fix, develop and release branches.  To upload and build the wheel I use [twine](https://pypi.org/project/twine/) which enables you to upload to PyPI in a secure/easy manner. Here is a sample `setup.py` I made for a project ([example setup.py](https://github.com/deephyper/deephyper/blob/master/setup.py)), it references required packages, command lines, version, twine for PyPI and other info. It works well for extras packages used in test or documentation.

rth:
I agree @Deathn0t . The choice of `setup.py` vs `setup.cfg` is orthogonal to this issue.  `setup.cfg` is indeed a bit nicer, but in the end I would say it doesn't really matter much.

[...]

PR #127, opened 2020-04-01
Improve versioning from git tags with setuptools_scm

rth:
One possible way to close https://github.com/icubam/icubam/issues/121 using https://github.com/pypa/setuptools_scm which determines version from git tags + commit hash.

This also adds `pyproject.toml` as a recommended way to using that package, since `setuptools_scm` now needs to be available when installing, and this ensures it. This part is not absolutely necessary, can be reverted.

The workflow is just to add tags from time to time, and the version [will be automatically generated](https://github.com/pypa/setuptools_scm#default-versioning-scheme), e.g.,
 - `0.1.8rc6` - if it's the tag exactly
 - `0.1.8rc4.dev0+g6dea7f7.d20200401` -  a few commits after the `0.1.8rc4` tag.

One limitation is that the tag needs to follow [semver](https://semver.org/), and incorrect tag scheme (for the last tag) will result in an installation error due to regex parsing such as,
```
      prefix, tail = re.match(r"(.*?)(\d+)$", version).groups()                                                                        
  AttributeError: 'NoneType' object has no attribute 'groups'
```
So for instance, `v0.1.8-deploy` is not a valid tag, while, `v0.1.8.deploy` might be OK (need to check).

Another limitation is that this package now needs to be installed before it can be used.


pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[build-system]
```
dulacarnold, line 1: Can these dependencies be simply defined in setup.py?

resources/config.toml
```diff
@@ -1,6 +1,5 @@
 # Keys
 default_locale='fr-fr'
-version='0.1.2'
```
dulacarnold, line 3: wooho!

rth:
For instance, the version currently installed in CI is `0.1.dev1+g80a4266` which should follow the following pattern (https://github.com/pypa/setuptools_scm#default-versioning-scheme),
```
{next_version}.dev{distance}+{scm letter}{revision hash}
```
the hash is different from that of the last commit because CI merged with upstream internally I think. It does work as expected for me locally.

If you don't like `setuptools_scm` and the reliance on git tags, as simpler solution could be just indicate the version in `setup.py`, and it all other places it would be retrieved from there (as done in this PR).

dulacarnold:
@rprimet this would require us to change the deployment rule on staging.  What do you think?  This deffinitelly satisfies the 'centralize our versioning to one place' need.

rprimet:
@dulacarnold I like it, I'll amend the versioning scheme in the deployment repository when this is merged

rth:
> what happens if someone deploys w/out git (e.g. docker, pip install from pypi or from tgz)?

 - if you install it from PyPi, it means the the `.tar.gz` was previously generated with `python setup.py sdist`, which currently for me generates `icubam-0.1.8rc7.dev3+g7c1fe35.tar.gz` and the version is internally included in the package metadata (`icubam.egg-info/PKG-INFO`). So it will work.
 - if you install from Github with git support,
   ```
   pip install git+https://github.com/rth/icubam.git@impr-version
   ```
   e.g. to install the branch in this PR, it will also work, since you have access to git history including tags there.
- If you install from Github without git support,
  ```
  https://github.com/rth/icubam/archive/impr-version.zip
  ```
  it will fail with the following error,
  ```
  LookupError: setuptools-scm was unable to detect version for '/tmp/pip-req-build-toutfj9_'.
  
  Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
  
  For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
  ```
  which is a reasonable and explicit error message.

So overall it should work.

pyproject.toml
rth: We could put them in `setup.py` but it's a legacy way, that can break in some conditions. Basically `setup.py` can define dependencies fine, but cannot (very well) specify what is needed to run `setup.py` itself (i.e. build dependencies) . It's a chicken an egg problem. `pyproject.toml` was desined for that. 

See blog post in https://twitter.com/brettsky/status/1245207783240110080 for more details.

rth: But we can remove it, no one is enthusiastic about it.

dulacarnold: LGTM, let's use pyproject.toml

PR #132, opened 2020-04-01
improve UX of update form (review & suggestions needed)

tgy:
new_ux


tgy:
Screenshot 2020-04-01 at 20 05 41

rth:
I'll try to test more locally, but I have a question/concern about not allowing submitting when some of the montonicity concerns failed.

Say a person arrives at this form and for some reason (e.g. the previous submission had a mistake) sees cumulated values higher than ones they want to report.

Currently the only course of actions possible is to keep them at previous value regardless.

Wouldn't it be better to show a big warning if they are decreasing (as it's done now), but nevertheless allow submitting data, even if it's not consistent with previous submission? Otherwise this might lead to a bias in collected data that would be even harder to correct in the end?

tgy:
very clever. i didn't think about that scenario! i will change the error to a warning. hopefully i made it very clear that the values are cumulative. thanks for that great insight @rth

rth:
Maybe let's wait for a second opinion before changing it? I might be missing something in how this form is typically filled.

tgy:
i think you're entirely right in your last comment @rth 

here's the new version. no disabling of the submit button now.

Screenshot 2020-04-01 at 22 13 19

rth:
> here's the new version. no disabling of the submit button now.

Looks good. Maybe repeat "décès (cummulés)" and "transferts (cummulés)" in the error message just to be more explicit about why this is an issue?

tgy:
Screenshot 2020-04-01 at 22 47 22

PR #138, opened 2020-04-01
User access token in external API calls 

olivierteboul:


dulacarnold:
LGTM, but need a token to test.

PR #139, opened 2020-04-01
BUG fix to_pandas for nested objs

tgy:


olivierteboul:
Thanks Valentin. However I would follow a different strategy.
First the function has little to do with pandas, since you only call the pd.DataFrame constructor at the end.
So what about having a function `def flatten(d: dict) -> dict` that flattens a dictionary. It can be defined in another module. Like `icubam.dict_utils.py` (does not exist yet)

Then for example `flatten({'k1': {'k2': 1, 'k3': 8}, 'k4': 6}) = {'k1_k2': 1, 'k1_k3': 8, 'k4': 6}`

Which mean that you could simply do:
```
def to_pandas(objs):
  return pd.DataFrame([dict_utils.flatten(obj.to_dict(max_depth=1)) for obj in objs])
```

what do you think?

tgy:
@olivierteboul makes sense

tgy:
@olivierteboul let me know

rth:
FYI, there is a [`json_normalize`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.json_normalize.html) in pandas that does this:
```py
>>> from pandas.io.json import json_normalize

>>> data = [{'id': 1, 'name': {'first': 'Coleen', 'last': 'Volk'}},
>>>        {'name': {'given': 'Mose', 'family': 'Regner'}},
>>>        {'id': 2, 'name': 'Faye Raker'}]

>>> json_normalize(data)
    id        name name.family name.first name.given name.last
0  1.0         NaN         NaN     Coleen        NaN      Volk
1  NaN         NaN      Regner        NaN       Mose       NaN
2  2.0  Faye Raker         NaN        NaN        NaN       NaN
```

(I also happened to re-implement something similar for another project before discovering this function).

tgy:
@rth you're a rock star. thanks.

icubam/www/handlers/db.py
```diff
@@ -1,4 +1,5 @@
 from absl import logging  # noqa: F401
+import pandas as pd
```
rth, line 2: Unused import

icubam/db/store.py
```diff
@@ -736,5 +736,7 @@ def create_store_for_sqlite_db(cfg) -> Store:
   return Store(engine, salt=cfg.DB_SALT)
 
 
-def to_pandas(objs) -> pd.DataFrame:
-  return pd.DataFrame([x.to_dict() for x in objs])
+def to_pandas(objs):
+  return pd.io.json.json_normalize(
```
olivierteboul, line 740: good job: from a complex function to a one-liner!! awesome

olivierteboul:
:clap: you could add test... but you can also merge.

rth:
Let's merge then. Just need to rebase @tgy . Thanks for the fix!

tgy:
i tested it locally and it works when pulling the data. should be ok

PR #141, opened 2020-04-01
Finishing up user creation and edition flow

fpquintao:


icubam/backoffice/handlers/users.py
```diff
@@ -39,50 +40,111 @@ def get(self):
       '{}?id={}'.format(UserHandler.ROUTE, self.user.user_id))
 
 
-
 class UserHandler(base.BaseHandler):
   ROUTE = "/user"
 
+  """Creates ICU DTOs based on ALL ICUs and those from the subset."""
+  def prepare_icus(self, subset_icus_ids):
```
olivierteboul, line 47: maybe call those selected_icus instead of subset ?

icubam/backoffice/handlers/users.py
```diff
@@ -39,50 +40,111 @@ def get(self):
...
+    icus = self.store.get_icus()
```
olivierteboul, line 48: there should be a slight difference between admin and managers. But let's have a todo here `TODO(olivier): restrict managers to regional icus`

icubam/backoffice/handlers/users.py
```diff
@@ -39,50 +40,111 @@ def get(self):
...
+    # Password is not required when editing the user.
+    password = self.get_body_argument("password", "")
+    if password:
+      user_dict["password_hash"] = self.store.get_password_hash(password)
+
+    uid = self.get_body_argument("user_id", None)
+    if uid:
+      user_dict["user_id"] = int(uid)
+
+    # We want to keep this "temp" user, because if saving fails, we need to
```
olivierteboul, line 107: cool

icubam/backoffice/handlers/users.py
```diff
@@ -39,50 +40,111 @@ def get(self):
...
+    for key in ["name", "email", "telephone"]:
+      user_dict[key] = self.get_body_argument(key, "")
+
+    # Password is not required when editing the user.
+    password = self.get_body_argument("password", "")
+    if password:
+      user_dict["password_hash"] = self.store.get_password_hash(password)
+
+    uid = self.get_body_argument("user_id", None)
+    if uid:
```
olivierteboul, line 104: better use the `is not None` syntax IMO, since other things could return False (0, "", etc.)

icubam/backoffice/handlers/users.py
```diff
@@ -39,50 +40,111 @@ def get(self):
...
 
+  """Creates ICU DTOs based on ALL ICUs and those from the subset."""
+  def prepare_icus(self, subset_icus_ids):
+    icus = self.store.get_icus()
+    output = []
+    for icu in icus:
+      dto = icu.to_dict()
+      dto["selected"] = icu.icu_id in subset_icus_ids
+      output.append(dto)
+    return sorted(output, key = lambda x: x['name'])
```
olivierteboul, line 54: nice

icubam/backoffice/handlers/users.py
```diff
@@ -39,50 +40,111 @@ def get(self):
...
-      user = self.db.get_user(userid)
-    # TODO(fred): load the ICUs to show in the form.
+      user = self.store.get_user(userid)
+
     user = user if user is not None else store.User()
-    self.render("user.html", icus=[], user=user, error="")
+    icus_dto = self.prepare_icus_for_user(user)
+    self.render(
+      "user.html", icus=icus_dto, user=user, error=False,
+      allow_admin=self.user.is_admin)
```
olivierteboul, line 71: you can remove `allow_admin`, and in the form simply user {{this_user.is_admin}}. I have overloaded self.render to always inject this_user=self.user :)

icubam/backoffice/templates/user.html
```diff
@@ -47,37 +68,105 @@ <h3 class="card-title">Create User</h3>
...
+             </div>
+             <!-- /.row -->
+           </div>
+           <!-- /.card-body -->
+       {% end %}
+       </div>
+       <!-- /.card -->
 
+
+       {% if allow_admin %}
```
olivierteboul, line 123: `{% if this_user.is_admin %}`

icubam/backoffice/handlers/users.py
fpquintao: sure, also renamed the methods to be more like "select_" instead of "prepare_"

icubam/backoffice/handlers/users.py
fpquintao: Added.

icubam/backoffice/handlers/users.py
fpquintao: ack

icubam/backoffice/handlers/users.py
fpquintao: hahahaha I see it now in BaseHandler.
G'enial.

icubam/backoffice/handlers/users.py
fpquintao: yeah, I thought about it after going to bed yesterday.

icubam/backoffice/handlers/users.py
fpquintao: ack

icubam/backoffice/handlers/users.py
fpquintao: there's a small little thing that I don't understand well with hidden fields (they come empty always, not None). I will check that later, so I had to keep the same check here.

olivierteboul: ok!

olivierteboul: probably because it is body argument. so there is no such thing as None. it s all strings

PR #145, opened 2020-04-02
adding icu pages in backoffice

olivierteboul:


icubam/backoffice/handlers/icus.py
```diff
@@ -0,0 +1,60 @@
...
+    if self.user.is_admin:
+      regions = self.store.get_regions()
+    if not self.user.is_admin:
+      regions = [e.region for e in self.user.managed_icus]
+    regions.sort(key=lambda r: r.name)
+    self.render("icu.html", icu=icu, regions=regions, error='')
+
+  @tornado.web.authenticated
+  def post(self):
+    icu_id = self.db.get_user(self.get_query_argument('id', None))
```
fpquintao, line 52: same here, perhaps I' missing something, but this should bet get_icu ?
the other question is: this should invoke self.store, not call self.db ???? Or it does not matter?

icubam/backoffice/handlers/icus.py
```diff
@@ -0,0 +1,60 @@
...
+    values = self.parse_from_body(store.ICU)
+    values["is_active"] = bool(values["is_active"])
+    if icu_id is None:
```
fpquintao, line 55: perhaps add a try/catch here around the call to the store, since you're not handling any validation of the data from the form? Then you return self.render(error="True", icu=values) at least and the server does not return a 500 in case the DB returns an exception.

icubam/backoffice/handlers/icus.py
```diff
@@ -0,0 +1,60 @@
...
+      self.db.add_icu(self.user.user_id, store.ICU(**values))
+    else:
+      self.db.update_icu(self.user.user_id, icu_id, values)
```
fpquintao, line 58: self.db or self.store?
(or it does not matter?)

icubam/backoffice/handlers/icus.py
```diff
@@ -0,0 +1,60 @@
...
+      "list.html", data=data, columns=columns, objtype='ICUs',
+      create_route=ICUHandler.ROUTE)
+
+
+class ICUHandler(base.BaseHandler):
+  ROUTE = "/icu"
+
+  @tornado.web.authenticated
+  def get(self):
+    icu = self.db.get_user(self.get_query_argument('id', None))
```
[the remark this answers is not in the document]
olivierteboul: yes!

icubam/backoffice/handlers/icus.py
olivierteboul: actually they  are the same (see base_server). removing the self.store one elsewhere

icubam/backoffice/handlers/icus.py
olivierteboul: let me add a todo.

icubam/backoffice/handlers/icus.py
olivierteboul: same again

PR #150, opened 2020-04-02
message server client and notification from the user handler

fpquintao:
- tested locally: it works
- added an option to start all the 3 servers at once


icubam/messaging/server.py
```diff
@@ -22,7 +23,13 @@ def __init__(self, config, port=8889):
     self.callbacks = [self.process]
 
   def make_app(self):
-    return tornado.web.Application(self.routes)
+    self.add_handler(
+        onoff.OnOffHandler, db = self.db, scheduler=self.scheduler)
+
+    # Only accepts request from same host
+    return tornado.web.Application([
+        (tornado.routing.HostMatches(r'(localhost|127\.0\.0\.1)'), self.routes),
```
dulacarnold, line 31: Maybe use double quotes here?

resources/config.toml
```diff
@@ -27,16 +27,24 @@ default_locale='fr-fr'
...
-    [scheduler.dev]
-    max_retries = 3
-    reminder_delay = 45
-    ping = ['12:30', '17:30', '14:12']
+  [messaging.dev]
+  PORT = 8889  # will be lower cased when reading.
+  base_url = "http://localhost:8889/"
+  [messaging.prod]
+  PORT = 8889  # will be lower cased when reading.
+  base_url = "https://test.icubam.net/"
```
dulacarnold, line 35: Probably best to keep localhost default here.   In general best to keep the icubam DNS out of the codebase commits.

icubam/messaging/client.py
```diff
@@ -0,0 +1,29 @@
...
+                   user_id: int,
+                   icu_ids: List[int],
+                   on: bool = True,
+                   delay: Optional[int] = None):
+    """Notify the scheduler that a user must be added / remove from the loop."""
+    if not icu_ids:
+      return logging.info('nothing to change. Aborting')
+
+    request = onoff.OnOffRequest(user_id, list(icu_ids), on, delay)
+    url = os.path.join(self.config.messaging.base_url,
```
fpquintao, line 26: I guess this can be set in the constructor as self.url = ....

olivierteboul: we might have other routes... then the url would be ambiguous i believe

PR #153, opened 2020-04-02
adding regions to BO

olivierteboul:


icubam/backoffice/handlers/regions.py
```diff
@@ -0,0 +1,53 @@
...
+import tornado.escape
+import tornado.web
+
+from icubam.backoffice.handlers import base
+from icubam.backoffice.handlers import home
+from icubam.db import store
+
+
+class ListRegionsHandler(base.BaseHandler):
+  ROUTE = "/list_regions"
```
rth, line 11: Maybe
```suggestion
  ROUTE = "/list-regions"
```

icubam/backoffice/handlers/regions.py
```diff
@@ -0,0 +1,53 @@
...
+
+  def prepare_for_table(self, region):
+    result = region.to_dict(max_depth=0)
+    for key in ['last_modified', 'icu']:
+      result.pop(key, None)
+    return self.format_list_item(result)
+
+  @tornado.web.authenticated
+  def get(self):
+    if self.user.is_admin:
```
rth, line 21: Isn't there a decorator to make endpoints only accessible by a given user? 

Otherwise we will write this often, which is also fine, just a comment.

icubam/backoffice/templates/base.html
```diff
@@ -115,6 +115,12 @@
               <p>Access Tokens</p>
             </a>
           </li>
+          <li class="nav-item">
+            <a href="/list_regions" class="nav-link">
```
rth, line 119:
```suggestion
            <a href="/regions" class="nav-link">
```

rth: I think the fact that `GET /regions` returns a list of regions is expected and there is no need for `list` in the URL.

olivierteboul: as much as i agree with you, all the other handlers are the same + there it is slightly less error prone. But in principle i would prefer just regions.

icubam/backoffice/handlers/regions.py
olivierteboul: Very good idea. I have added an AdminHandler now

rth:
There is a CI failure, otherwise LGTM.

olivierteboul:
thanks Roman, i have updated the PR

PR #155, opened 2020-04-02
TST Add some code coverage

VincentRouvreau:


icubam/messaging/sms_sender_test.py
```diff
@@ -21,6 +21,13 @@ def test_from_string(self):
       sms_sender.get(self.config, 'Fake'), sms_sender.FakeSender)
     self.assertIsInstance(
       sms_sender.get(self.config, 'fake'), sms_sender.FakeSender)
+    with self.assertRaises(ValueError):
```
rth, line 24: might be good to use `assertRaisesRegex` with part of the error message, to be more specific.

icubam/messaging/sms_sender_test.py
```diff
@@ -21,6 +21,13 @@ def test_from_string(self):
...
+      sms_sender.get(self.config, 'dummy_sms_sender_name')
+
 
   def test_from_config(self):
     self.assertIsInstance(sms_sender.get(self.config), sms_sender.FakeSender)
+
+  def test_send_from_fake_sender(self):
+    ms = sms_sender.get(self.config, 'Fake')
+    ms.send("Fake phone number", "Hello from tests")
```
rth, line 33: Maybe just add an inline comment explaining what this does: i.e. I hope this actually won't send a SMS if the token is provided..

dulacarnold: Fake sender is safe.

VincentRouvreau: I add a comment to make it clearer

icubam/messaging/sms_sender_test.py
VincentRouvreau: Good idea !

Issue #154, opened 2020-04-02
Boheh plot server


rth:
It looks like the decision was taken to go with bokeh for plots. I had a quick look on what would be needed to make interactive visualization e.g. have a drop down of regions and then update the plot when a region is selected. Or change the variable that's plotted by changing a dropdown.

In bokeh, it's simple to add JS callbacks, but I don't immediately see how we could make it work for the interactivity required if we need to run some python code to get the result. We could add a news endpoint that returns data with JSON but then it's a lot of JS coding.

Another alternative would be to run Bokeh server in the backend. It's possible to start it alongside the Tornado app (cf [docs](https://docs.bokeh.org/en/latest/docs/user_guide/server.html#embedding-bokeh-server-as-a-library) and the Tornado example linked there), but that's doesn't look that simple to me. Maybe I'm too tired and not familiar enough with Tornado / Bokeh server. WDYT @olivierteboul ?

Personally I have used more dash/plotly or [voila]( voila
) for interactive dashboards so far.

It's likely possible to make the bokeh server working embedded in the backoffice app, but I don't have a very good understanding how much work it would be. Otherwise it would mean that we would be limited to non-interactive visualizations.

cc @Deathn0t

rth:
To be more precise, I'm not saying that bokeh is a bad choice. Eeach of bokeh, dash, voila are quite straightforward to make an interactive dashboards with if one uses they server in development. For use in production, if we don't want to have yet another server process running, it's a bit more work (but likely manageable).

rth:
Closing as no longer relevant with the analytics server implementation in https://github.com/icubam/icubam/pull/407

PR #159, opened 2020-04-03
Improve /operational-dasboard

rth:
Improves the operational dashboard (previously was known as  `/plot`).

This endpoint is expected to show the situation at a given time. 

WIP because I still want to add a table with key indicators on the left. 
There is a selector by region, because of https://github.com/icubam/icubam/issues/154 for now the full page is reloaded when region is changed.

If interested I can send you screenshots by DM.


icubam/backoffice/handlers/operational_dashboard.py
```diff
@@ -0,0 +1,151 @@
...
+    current_region_name = "Toutes les régions"
+    if arg_region is not None and arg_region.isdigit():
+      try:
+        current_region = self.db.get_region(int(arg_region))
+        current_region_name = current_region.name
+      except Exception:
+        logging.warning(
+          f"Invalid query argument: region={arg_region} "
+          f"Falling back to all regions."
+        )
```
rth, line 34: I know I should add tests for this path, but for I haven't managed to make it work (user needed to be monkeypatched, and I was likely missing something).

icubam/backoffice/templates/operational-dashboard.html
```diff
@@ -0,0 +1,85 @@
...
+    <div class="row">
+        <div class="col-sm-5 dash-metrics">
+	   {% for row in metrics_layout %}
+	   <div class="row"> 
+	      {% for metric in row %}
+              <div class="col-sm-4">
+                 <div class="card">
+                    <div class="card-body">
+		       <h3 class="card-title">{{ metric['value'] }}</h3>
+		       <p class="card-text">{{ metric['label'] }}</p>
```
rth, line 48: Would you know a formatter for jinja2 files by any chance? Struggling with tabs here :/

olivierteboul: yeah. it s bit messy but no worries.

icubam/backoffice/handlers/operational_dashboard.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class OperationalDashHandler(base.BaseHandler):
+  ROUTE = '/operational-dashboard'
+
+  @tornado.web.authenticated
+  async def get(self):
+    """Serves a page with a table gathering current bedcount data with some extra information."""
+    arg_region = self.get_query_argument('region', default=None)
+    current_user = self.get_current_user()
```
olivierteboul, line 24: You actually have access to `self.user`

icubam/backoffice/handlers/operational_dashboard.py
olivierteboul: you can just do like the other with mock.patch. Then you can try different requests  with different regions..

Check  backoffice/server_test.py

icubam/backoffice/handlers/operational_dashboard.py
```diff
@@ -0,0 +1,151 @@
...
+        current_region_name = current_region.name
+      except Exception:
+        logging.warning(
+          f"Invalid query argument: region={arg_region} "
+          f"Falling back to all regions."
+        )
+
+    figures = []
+
+    bed_counts = self.db.get_visible_bed_counts_for_user(current_user.user_id)
```
olivierteboul, line 40: use self.user instead of current_user

icubam/backoffice/handlers/operational_dashboard.py
```diff
@@ -0,0 +1,151 @@
...
+
+    bed_counts = to_pandas(bed_counts)
+
+    if current_region is not None:
+      mask = bed_counts['icu_region_id'] == current_region.region_id
```
olivierteboul, line 45: i would check that,  `if not self.user.is_admin`  then we should make sure that the requested region_id is part of the region_ids of the managed_icus.

olivierteboul: we do not want the a user to have access to regions he does not belong

icubam/backoffice/handlers/operational_dashboard.py
```diff
@@ -0,0 +1,153 @@
...
+
+from icubam.db.store import to_pandas
+from icubam.backoffice.handlers import base
+
+
+class OperationalDashHandler(base.BaseHandler):
+  ROUTE = '/operational-dashboard'
+
+  @tornado.web.authenticated
+  async def get(self):
```
olivierteboul, line 21: it is a very long function so i would split the logic of building the plot knowing that you have the correct dataFrame in another function and call it here. it would be more readable.

rth:
OK I have made the following changes,
 - split main method into multiple functions to help readability
 - Added localization. @olivierteboul Do you know what I need to do to see this page in french? I am currently starting the server as `python scripts/run_server.py --config=config.toml --server=backoffice` where   `config.toml` does have `default_locale='fr-fr'` (but my browser has English locale by default). Haven't localized the plot labels (they are all in French) for now.
 - Skip the test for `?region=<id>` queries, that fail for reasons unrelated to this PR (something about how we monkeypatch user)
 - Made aggregation by department instead of city, as requested so it's less sensitive.
 - Made this endpoint only accessible by admin. Until we figure out permissions. Essentially this dashboard shows information aggregated by region / department. If we only display information from ICUs that a given user has access to this aggregated information will be wrong. Also I still think filtering should be done at the level of DB query not handler.

@olivierteboul I would like to merge this as is for now, then iteratively improve it. In any case this is an admin only page for now.

PR #160, opened 2020-04-03
Fixed timestamp offset bug.

dulacarnold:


icubam/time_utils.py
```diff
@@ -26,7 +21,9 @@ def time_ago(ts, ts_reference=None) -> Tuple[int, str]:
   if ts is None:
     return -1, 'never'
 
-  ts_reference = time.time() if ts_reference is None else ts_reference
+  ts_reference = int(
+    datetime.datetime.utcnow().timestamp()
```
dulacarnold, line 25: This is the actual change.

PR #161, opened 2020-04-03
supporting timezone

olivierteboul:
a bit of an annoying PR, but i think it will solve the problem. I have to double check though

### DIFF
--- a/icubam/time_utils.py
+++ b/icubam/time_utils.py
@@ -1,9 +1,16 @@
-from typing import Tuple
-import functools
-import time
 import datetime
+import functools
+import pytz
+import time
+from typing import Tuple
+
 
 UNITS = [(86400, 'day'), (3600, 'hour'), (60, 'minute'), (1, 'second')]
+
+
+def utcnow() -> int:
+  """The current timestamp in utc."""
+  return int(datetime.datetime.utcnow().timestamp())
 
 
 def time_ago(ts, ts_reference=None) -> Tuple[int, str]:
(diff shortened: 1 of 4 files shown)


icubam/time_utils.py
```diff
@@ -54,29 +59,43 @@ def parse_hour(hour, sep=':') -> Tuple[str, str]:
...
    ts: an optional timestamp. If none, use the current timestamp.
+   tz: a pytz.timezone to parse the the hours
   """
   if not hours:
     return None
 
-  ts = int(time.time()) if ts is None else ts
-  now = datetime.datetime.fromtimestamp(ts)
+  ts = utcnow() if ts is None else ts
+  now = to_utc(datetime.datetime.fromtimestamp(ts))
```
dulacarnold, line 86: This should already be in UTC if it comes from the DB.  Not sure what to_utc assumes.

olivierteboul: it s a bit more subtle indeed. what comes from the database is "naive", meaning, if i understand correctly, that it is utc based but it is not specified it is utc. In order to be able to compare two dates, it seems like if one is naive and one is localized, it fails. So here, the only thing this `to_utc` does is to say: specify the timezone to be utc, so that later, we will take the candidates datetimes (parsed from config) translate them into utc and compare to find the next one

olivierteboul:
I am leaving this on hold since it s a bit complex. i ll come back to it later...

olivierteboul:
Closing for now

PR #162, opened 2020-04-04
many fixes in backoffice

olivierteboul:
- add hidden object_id in form
- and try catch in all forms + show error
- add cancel button


icubam/backoffice/handlers/icus.py
```diff
@@ -31,29 +31,37 @@ def get(self):
...
-    if icu is None:
-      self.db.add_icu(self.user.user_id, store.ICU(**values))
-    else:
-      self.db.update_icu(self.user.user_id, icu.icu_id, values)
+    values["is_active"] = values["is_active"] == 'True'
+    id_key = 'icu_id'
+    icu_id = values.pop(id_key, '')
+    try:
+      if not icu_id:
+        self.db.add_icu(self.user.user_id, store.ICU(**values))
```
fpquintao, line 59: not really needed but just in case we want to keep the user in the same page instead of redirecting in line 67 (if people ask us because they want to keep editing or double check what they had done), this will be much safer if we assign the new icu_id:
icu_id = self.db.add_icu(self.user.user_id, store.ICU(**values))

icubam/backoffice/handlers/icus.py
```diff
@@ -31,29 +31,37 @@ def get(self):
...
+  @tornado.web.authenticated
+  def post(self):
     values = self.parse_from_body(store.ICU)
-    values["is_active"] = bool(values["is_active"])
-    # TODO(olivier): try catch error here.
-    if icu is None:
-      self.db.add_icu(self.user.user_id, store.ICU(**values))
-    else:
-      self.db.update_icu(self.user.user_id, icu.icu_id, values)
+    values["is_active"] = values["is_active"] == 'True'
```
fpquintao, line 54: make sure this is not "on"
(apparently it depends on the checkbox component you use, but the one used in the users page gives:
empty string when false
"on" when true

I usually do print(self.request.body) to debug how the values arrive in the handler.

icubam/backoffice/handlers/tokens.py
```diff
@@ -29,20 +30,26 @@ def get(self):
...
-      self.db.add_external_client(self.user.user_id, incoming_user)
-    else:
-      c_id = user.external_client_id
-      self.db.update_external_client(self.user.user_id, c_id, values)
+    values = self.parse_from_body(store.ExternalClient)
+    id_key = 'external_client_id'
+    token_id = values.pop(id_key, '')
+    try:
+      if not token_id:
+        self.db.add_external_client(
```
fpquintao, line 46: same here about the regiontoken_id

icubam/backoffice/handlers/icus.py
olivierteboul: good point. but because of line 44 and the component has a value in icu.html, it comes True / False as string

icubam/backoffice/handlers/icus.py
olivierteboul: done

icubam/backoffice/handlers/tokens.py
olivierteboul: yep

PR #165, opened 2020-04-04
Use scoped session in DB store.

sgirgin:


rth:
The `SchedulerTestCase.test_schedule_all` is actually ignored because there are two identically named such methods. Once naming is resolved in https://github.com/icubam/icubam/pull/167/  it fails with,
```
       self.scheduler.schedule_all()
      # We are not sure about what is the db. But at least it should send to the
      # newly built users.
      # TODO(olivier): do better here
>     self.assertGreater(len(self.scheduler.timeouts), len(names) + 1)
E     AssertionError: 5 not greater than 5
```

PR #166, opened 2020-04-04
Adding a ping on others servers in backoffice

olivierteboul:
This PR adds a systematic ping to the www and message servers from the backoffice and always gives the status of those two in the nav bar (for admins only)


olivierteboul:
image

rth:
Nice! Just to confirm: so each of those handlers now need to be async, otherwise for each request we will be waiting for the pinged server to respond?

olivierteboul:
i don t think it would hang, but i am not sure what would happen. I can give it a try to double check. But having async handlers is not a big deal. on the contrary you get more than you loose

icubam/backoffice/handlers/base.py
```diff
@@ -13,9 +14,25 @@ def initialize(self, config, db):
...
+        result[key] = {'status': resp.code == 200, 'since': resp.body}
+      except:
+        result[key] = {'status': False, 'since': '-'}
+        continue
+    return result
 
-  def render(self, path, **kwargs):
-    super().render(path, this_user=self.user, **kwargs)
+  async def render(self, path, **kwargs):
+    pings = await self.ping() if self.user and self.user.is_admin else {}
```
sgirgin, line 34: Should this be handled by a periodic callback instead? It is easy to define one in tornado it seems and will be better than pinging at each request.

olivierteboul: sure. let me give it a try

sgirgin: Tentatively approving as discussed. You can change it later.

olivierteboul: done !

PR #171, opened 2020-04-05
Added store factory and made sessions request scoped.

sgirgin:
This is a synthetic change and currently a no-op. It will allow us to move to request scoped sessions. You may possibly skip the tests, except the end of the file where we add StoreWithSessionTest.


sgirgin:
I just realized that there may be a nicer and less intrusive way of doing this. Let me give it a try and I'll merge this one if it doesn't work.

PR #173, opened 2020-04-05
Two BO fixes: rendering active checkbox in manager mode and saving a manager

olivierteboul:


icubam/backoffice/static/user.js
```diff
@@ -1,7 +1,13 @@
 $(function () {
   $('.select2bs4').select2()
 
-  function isAdmin () { return $('#is_admin')[0].checked }
+  function isAdmin () {
+    let admin_box = $('#is_admin')
+    if (admin_box !== undefined && $('#is_admin')[0] !== undefined) {
```
sgirgin, line 6: You may use admin_box in the second condition and return statement.

PR #174, opened 2020-04-05
bug fix in checkbox. fred was right :)

olivierteboul:


icubam/backoffice/handlers/users.py
```diff
@@ -75,6 +75,7 @@ def get_options(self):
     return self.db.get_managed_icus(self.user.user_id)
 
   def prepare_for_display(self, user: store.User):
+    print("--->", user.is_active, user.is_admin)
```
sgirgin, line 78: Remove this one?

PR #179, opened 2020-04-06
better individual lists

olivierteboul:
Put more relevant content in ICUs, Users, etc.

### DIFF
--- a/icubam/backoffice/handlers/messages.py
+++ b/icubam/backoffice/handlers/messages.py
@@ -3,8 +3,7 @@ import logging
 import tornado.escape
 import tornado.web
 
-from icubam.backoffice.handlers import base
-from icubam.backoffice.handlers import home
+from icubam.backoffice.handlers import base, home, icus, users
 from icubam.db import store
 from icubam.messaging import client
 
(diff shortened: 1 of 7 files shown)


icubam/backoffice/handlers/messages.py
```diff
@@ -18,9 +17,34 @@ def initialize(self):
...
+        'link': f'{icus.ICUHandler.ROUTE}?id={msg["icu_id"]}'
+      },
+    ]
+    msg_dict = {}
+    msg_dict['telephone'] = msg['phone']
+    msg_dict['scheduled'] ='{0:%Y/%m/%d at %H:%M:%S}'.format(
+        datetime.datetime.fromtimestamp(msg['when']))
+    msg_dict['attempts'] = msg['attempts']
+    msg_dict['first sent'] = 'not yet'
+    if msg['first_sent'] is not None:
```
fpquintao, line 38: what's the difference bewteen "first_sent" and "first sent"? can we use the same key format?
(I feel like a key with space is a bit weird and error prone depending on how things get escaped etc)

PR #188, opened 2020-04-07
fixing user save in manager mode

olivierteboul:
It is a fix with a TODO since i believe a good query should fix the problem

### DIFF
--- a/icubam/db/store.py
+++ b/icubam/db/store.py
@@ -384,7 +384,7 @@ class Store(object):
     Returns:
       ID of the user.
     """
-    self._session.add(user)icu_ids
+    self._session.add(user)
     self._session.commit()
     return user.user_id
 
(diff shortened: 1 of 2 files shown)


icubam/db/store.py
```diff
@@ -463,11 +463,25 @@ def manages_user(self, manager_user_id: int, user_id: int) -> bool:
     """Returns true if the manager user can manage the user."""
     if self.is_admin(manager_user_id) or manager_user_id == user_id:
       return True
-    return self._session.query(
+
+    # TODO(olivier): for some reason, this query sometimes returns False
+    # while it should not.
+    result = self._session.query(
```
sgirgin, line 469: Do you have an example? What we do below is essentially not different from this, but in a more convoluted way and I'm not sure if it will solve the issue.

PR #191, opened 2020-04-07
Remove access to user info from API calls.

dulacarnold:


tgy:
Very good job @dulacarnold! Keep them comin'

PR #198, opened 2020-04-08
Some cleanup, and updated upload endpoint for API auth.

dulacarnold:


icubam/www/handlers/upload_csv.py
```diff
@@ -2,24 +2,36 @@
 import tornado.web
 import logging
 import time
+from icubam.www.handlers import base
 
-@tornado.web.authenticated
-class UploadHandler(tornado.web.RequestHandler):
+
+class UploadHandler(base.BaseHandler): #tornado.web.RequestHandler):
```
olivierteboul, line 8: remove comment

icubam/www/handlers/upload_csv.py
```diff
@@ -2,24 +2,36 @@
...
   """Accept file uploads over POST."""
 
   ROUTE = "/upload_csv"
+  API_COOKIE = 'api'
+
+  def get_current_user(self):
```
olivierteboul, line 14: i think it would be better to add a APIKeyHandler that inherits from BaseHandler and that redefines this exact function and then use it for DB, Map, and Upload. This way u do not need to duplicate code

icubam/www/handlers/upload_csv.py
```diff
@@ -2,24 +2,36 @@
...
     self.upload_path = upload_path
+    print(db_factory)
+    super().initialize(config, db_factory)
 
+  @tornado.web.authenticated
   def post(self):
     file = self.request.files["file"][0]
     file_path = self.gen_path(file["filename"])
-
+    print(self.request)
```
olivierteboul, line 34: remove

icubam/www/handlers/upload_csv.py
```diff
@@ -2,24 +2,36 @@
...
+    return self.db.auth_external_client(key)
 
   def gen_path(self, f_name):
     time_str = time.strftime("%Y-%m-%d-%H:%M:%S")
     return os.path.join(self.upload_path, f"{time_str}-{f_name}")
 
-  def initialize(self, upload_path):
+  def initialize(self, upload_path, config, db_factory):
     self.upload_path = upload_path
+    print(db_factory)
```
olivierteboul, line 27: remove

dulacarnold: ugh sorry

icubam/www/handlers/upload_csv.py
dulacarnold: Agreed, let's do this on a separate PR.

PR #203, opened 2020-04-08
fixing static routes

olivierteboul:
and removed some prints

### DIFF
--- a/icubam/backoffice/templates/map.html
+++ b/icubam/backoffice/templates/map.html
@@ -2,9 +2,9 @@
 
 {% block links %}
 <link rel="stylesheet" href="static/map.css?v={{version}}">
-<link rel="stylesheet" href="/www/static/popup.css?v={{version}}">
-<script type="text/javascript" src="/www/static/popup.js?v={{version}}"></script>
-<script type="text/javascript" src="/www/static/map.js?v={{version}}"></script>
+<link rel="stylesheet" href="www/static/popup.css?v={{version}}">
+<script type="text/javascript" src="www/static/popup.js?v={{version}}"></script>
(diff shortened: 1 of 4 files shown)


icubam/backoffice/templates/map.html
```diff
@@ -2,9 +2,9 @@
 
 {% block links %}
 <link rel="stylesheet" href="static/map.css?v={{version}}">
-<link rel="stylesheet" href="/www/static/popup.css?v={{version}}">
-<script type="text/javascript" src="/www/static/popup.js?v={{version}}"></script>
-<script type="text/javascript" src="/www/static/map.js?v={{version}}"></script>
+<link rel="stylesheet" href="www/static/popup.css?v={{version}}">
```
rth, line 5: I thin the leading / is expected no?

dulacarnold: I think the nuance is between frontend and backend files, but agree it's not totally clear. @olivierteboul ?

rth: I mean aren't absolute URLs always better? Otherwise this will depend from where it rendered.

olivierteboul: it is actually not expected
if it starts with '/' then the path will be /www/static/...
if not it will depend on the current route. If it is icubam.net/bla/blo/map.html then it will be icubam.net/bla/blo/www/static...

dulacarnold: We /could/ explicitly add the backoffice route to the url and make it rooted in /, but I'm ok with this design choice.  @rth we run the backoffice in a redirect of a subdirectory fyi

PR #204, opened 2020-04-08
Missing package

pgraverdy:
Adds missing datclasses requirement

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ twilio==6.37.0
 click==7.1.1
 sqlalchemy==1.3.15
 bokeh==2.0.1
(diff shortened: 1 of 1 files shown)


pgraverdy:
Sorry, went a bit too quikly on that PR. Will fix first.

rth:
I think we are Python 3.7+ where https://docs.python.org/3.7/library/dataclasses.html is part of the stdlib?

dulacarnold:
I can confirm we are not compatible with python_version < 3.7.

dulacarnold:
@pgraverdy are you able to build w/out this if you use python>3.7?

PR #205, opened 2020-04-08
Added option to not wipe bed info.  Also reformatted to follow yapf config.

dulacarnold:


scripts/wipe_sensitive_info_db.py
```diff
@@ -10,64 +10,64 @@
...
-    for alteration in ALTERATIONS.items():
-        affectations = ",".join([f"{k} = {v}" for (k, v) in alteration[1]])
-        query = f"UPDATE {alteration[0]} SET {affectations};"
-        print(query)
-        conn.execute(query)
-    conn.commit()
-    conn.close()
+  conn = sqlite3.connect(path)
+  cur = conn.cursor()
+  if FLAGS.keep_beds:
```
rprimet, line 47: Is there a `not` missing here?

scripts/wipe_sensitive_info_db.py
```diff
@@ -10,64 +10,64 @@
...
+    ALTERATIONS["bed_counts"].extend([("n_covid_free", 2), ("n_covid_occ", 4)])
+  for alteration in ALTERATIONS.items():
+    affectations = ",".join([f"{k} = {v}" for (k, v) in alteration[1]])
+    query = f"UPDATE {alteration[0]} SET {affectations};"
+    print(query)
```
rprimet, line 52: The print statement should also probably be in a `if not FLAGS.quiet:` guard but maybe that's a separate PR?

scripts/wipe_sensitive_info_db.py
dulacarnold: completely, good catch

scripts/wipe_sensitive_info_db.py
dulacarnold: Done.

PR #207, opened 2020-04-08
Sanitized upload name some.

dulacarnold:


icubam/www/handlers/upload_csv.py
```diff
@@ -20,7 +20,17 @@ def get_current_user(self):
...
+    f_name = os.path.split(f_name)[1]
+    f_name, f_ext = os.path.splitext(f_name)
+    # Clean out non-alphanumerics (need to split ext first or we lose the '.'):
+    f_name, f_ext = [
+      "".join(c
+              for c in f_name
+              if c.isalnum())[:64]
+      for f_name in [f_name, f_ext]
+    ]
+    return os.path.join(self.upload_path, f"{time_str}-{f_name}.{f_ext}")
```
rth, line 33: A few unit tests wouldn't hurt to ensure the added functionality is working as expected. Can't this be done with a regex, something like?
```py
f_name = re.sub(r'\W+', '', f_name)
```

BTW, for path manipulations [pathlib](https://docs.python.org/3/library/pathlib.html)  is quite nice.

dulacarnold: Ok, I think os.path is sufficient here but will deff. give pathlib a good look.

icubam/www/handlers/upload_csv.py
```diff
@@ -1,10 +1,26 @@
...
 import time
 from icubam.www.handlers import base
 
 
+def clean_path(f_name):
+  time_str = time.strftime("%Y-%m-%d-%H:%M:%S")
+  # Sanitize the name some:
+  f_name = f_name.lower()
+  f_name = os.path.normpath(f_name)
+  f_name = os.path.split(f_name)[1]
```
olivierteboul, line 14: are we sure there is a 2nd element ?  should we prevent from raising error here?

icubam/www/handlers/upload_csv.py
```diff
@@ -18,18 +34,18 @@ def get_current_user(self):
...
   def initialize(self, upload_path, config, db_factory):
     self.upload_path = upload_path
     super().initialize(config, db_factory)
 
   @tornado.web.authenticated
   def post(self):
     file = self.request.files["file"][0]
-    file_path = self.gen_path(file["filename"])
+    file_path = clean_path(file["filename"])
+    os.path.join(
```
olivierteboul, line 45: you can os.path.join(a, b, c) instead of calling twice os.path.join

icubam/www/handlers/upload_csv.py
dulacarnold: path.split always returns two elements even if it has an empty [0].

PR #208, opened 2020-04-08
Optionally render extra plots in the /operational-dashboard

rth:
This adds an `extra_plots_dir` config parameter to `config.toml`. When it points to an existing dir, any png figure in that folder will be rendered at the end of `/operational-dashboard` page.

This is fully optional, if the `extra_plots_dir` is a non existing path or a folder with no `.png`, the previous behavour is unchanged.

TODO:
 - [ ] add a few tests
 - [ ] document this feature

Will add a separate PR to re-generate these figures periodically.

### DIFF
--- a/icubam/backoffice/server.py
+++ b/icubam/backoffice/server.py
@@ -78,6 +78,13 @@ class BackOfficeServer(base_server.BaseServer):
     self.add_handler(messages.ListMessagesHandler)
     self.add_handler(maps.MapsHandler)
 
+    if os.path.isdir(self.config.backoffice.extra_plots_dir):
+      route = os.path.join("/", self.root, r'static/extra-plots/(.*)')
+      self.routes.append(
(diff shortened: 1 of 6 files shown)


icubam/backoffice/server.py
```diff
@@ -78,6 +78,13 @@ def make_routes(self, path):
     self.add_handler(messages.ListMessagesHandler)
     self.add_handler(maps.MapsHandler)
 
+    if os.path.isdir(self.config.backoffice.extra_plots_dir):
+      route = os.path.join("/", self.root, r'static/extra-plots/(.*)')
```
dulacarnold, line 82: Were you able to check this works with self.root != ''?

rth: Yes, locally for me paths looks like `/bo/static/extra-plots/lineplot_beds_grand_est.png` where root is `/bo`

PR #213, opened 2020-04-09
Wipe unit tests

rprimet:
Add unit tests for the sensitive info wipe script.

Also refactor slightly some scripts to be shells around library functions


icubam/db/test_wipe.py
```diff
@@ -0,0 +1,33 @@
+from absl.testing import absltest
+from itertools import repeat
+from sqlalchemy import create_engine
+from icubam.db.fake import populate_store_fake
+from icubam.db.wipe import wipe_db
+from icubam.db.store import Store, StoreFactory
+
+class WipeTest(absltest.TestCase):
+    def setUp(self):
```
olivierteboul, line 9: you are using 4 spaces. please stick to 2

icubam/db/test_wipe.py
```diff
@@ -0,0 +1,33 @@
...
+      self.engine = create_engine("sqlite:///:memory:", echo=False)
+      store_factory = StoreFactory(self.engine)
+      self.store = store_factory.create()
+      populate_store_fake(self.store)
+
```
olivierteboul, line 14: remove blank line

icubam/db/test_wipe.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def test_wipe(self):
+        with self.engine.connect() as conn:
+            wipe_db(conn.connection, keep_beds=False)
+            users = list(self.store.get_users())
+            self.assertItemsEqual([user.name for user in users],
+                                  repeat("Jean Dumont", len(users)))
```
olivierteboul, line 21: i d rather keep the module, i think it s easier to understand when the function comes from: `itertools.repeat(...`

scripts/populate_db_fake.py
```diff
@@ -16,68 +16,7 @@ def main(unused_argv):
...
-
-  add_icu('Beaujon', '93', 'Bobigny', 48.90833333333333, 2.310277777777777,
-          'test_tel', 5, 6, 12, 200, 34, 7, 1)
-  icu_id = add_icu('Bicetre', '93', 'Kremelin-Bicetre', 48.81,
-                   2.353888888888889, 'test_tel', 9, 2, 12, 200, 34, 44, 1)
-  store.add_user_to_icu(
-      admin_user_id, icu_id,
-      User(name='user1', telephone='+336666666', description='desc1'))
-
+  populate_store_fake(store)
```
olivierteboul, line 19: good move!

scripts/wipe_sensitive_info_db.py
```diff
@@ -68,7 +35,7 @@ def main(argv):
     ):
       sys.exit(0)
 
-  wipe_db(argv[1])
+  wipe_db_path(argv[1], FLAGS.keep_beds)
```
olivierteboul, line 38: i would rather use FLAGS for both arguments, i think it s more readable and more consistent

scripts/wipe_sensitive_info_db.py
```diff
@@ -68,7 +35,7 @@ def main(argv):
     ):
       sys.exit(0)
```
olivierteboul, line 36: no need to sys.exit(0) inside a function, you can just return.

icubam/db/wipe.py
```diff
@@ -0,0 +1,52 @@
...
+  -------
+  None
+  """
+  cur = conn.cursor()
+  if not keep_beds:
+    ALTERATIONS["bed_counts"].extend([("n_covid_free", 2), ("n_covid_occ", 4)])
+  for alteration in ALTERATIONS.items():
+    affectations = ",".join([f"{k} = {v}" for (k, v) in alteration[1]])
+    query = f"UPDATE {alteration[0]} SET {affectations};"
+    print(query)
```
olivierteboul, line 47: remove print?

scripts/wipe_sensitive_info_db.py
rprimet: Hmmm OK but the filename is non-optional, whereas `keep_beds` is optional and therefore a flag?

icubam/db/test_wipe.py
rprimet: Will do, thanks!

icubam/db/test_wipe.py
rprimet: Yup seems clearer that way, thanks

scripts/wipe_sensitive_info_db.py
rprimet: OK, I thought it'd be more consistent with the `sys.exit(-1) above`

scripts/wipe_sensitive_info_db.py
dulacarnold: You can make the flag required with absl.  Agreed the flags should generally be kept optional according to argparse/general unix culture, but I also tend to like explicit flags vs. positional arguments.

scripts/wipe_sensitive_info_db.py
dulacarnold: Ok, if you are leveraging posix return codes this makes sense to me.

scripts/wipe_sensitive_info_db.py
rprimet: @dulacarnold @olivierteboul OK, filename is now a required flag

scripts/wipe_sensitive_info_db.py
rprimet: The sys.exit(-1) was rendered unnecessary by moving to an all-flags solution so I changed that as well to match your suggestion, thanks!

PR #214, opened 2020-04-09
Import csv

Duez:
ajout d'un module d'export/import de fichier .csv (db/csv.py) au format du google doc
plus quelques scripts stand alone(scripts/import_csv scripts/export_csv)

### DIFF
--- a/icubam/db/csv.py
+++ b/icubam/db/csv.py
@@ -53,31 +53,6 @@ class CSV:
 				# create ICU
 				icu_id = self.store.add_icu(admin_user_id, ICU(**icu_info))
 				print("IMPORT CSV : create ICU " + row[header.index("icu_name")])
-				'''
-				# create ICU default user
-				user_info = {
-					"name":				row[header.index("icu_name")] + "_user",
-					"telephone":	row[header.index("telephone")],
-					"email":			"a@bc.org",
-					"is_active":	True,
-					"is_admin":		False,
-				}
-				user_id = self.store.add_user_to_icu(admin_user_id ,icu_id, User(**user_info))
-
-				# update bed_count
-				bed_info = {
-					"icu_id": icu_id,
-					"n_covid_occ": 0,
-					"n_covid_free": 0,
-					"n_ncovid_occ": 0,
-					"n_ncovid_free": 0,
-					"n_covid_deaths": 0,
-					"n_covid_healed": 0,
-					"n_covid_refused": 0,
-					"n_covid_transfered": 0,
-				}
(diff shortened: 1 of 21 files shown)


icubam/db/csv.py
```diff
@@ -0,0 +1,132 @@
...
+
+	def __init__(self, db):
+		self.store = db
+
+	def get_default_admin(self):
+		# create default admin
+		return self.store.add_default_admin()
+
+	def import_icus(self,admin_user_id: int, csv_file_path:str, forceUpdate=False):
+		csv_data = csv.reader(open(csv_file_path))
```
rth, line 15: any reason not to use pandas.read_csv? That would generally be faster for processing that for loops over rows.

icubam/db/csv.py
```diff
@@ -0,0 +1,132 @@
...
+
+class CSV:
+	""" """
+
+	def __init__(self, db):
+		self.store = db
+
+	def get_default_admin(self):
+		# create default admin
+		return self.store.add_default_admin()
```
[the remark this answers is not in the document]
dulacarnold: Yes, there is a .style.yapf file.  Also we are on a 2-spaces indentation standard.

icubam/db/csv.py
dulacarnold: +1 for a pandas-based solution for maintainability down the line.

icubam/db/csv.py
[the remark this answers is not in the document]
tgy: happy to see it now serves a purpose.

icubam/db/csv.py
tgy: plus we already have pandas as a dependency

icubam/db/csv.py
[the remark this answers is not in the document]
Duez: this should be ok now

icubam/db/csv.py
Duez: i don't know panda(to be honest i don't know too much of python either) I will give it a look

Duez:
> Thanks! If would be necessary to add unit tests for the added functionality. e.g. under `icubam/db/csv_test.py` (see other test files in the repo)

added some tests

icubam/db/csv.py
tgy: simply use `pd.read_csv(csv_file_path)` here instead

tgy: and `import pandas as pd`

Duez: Done, it was a bit more than just a line, since iterating trough dataframe is done differently but in the end it's cleaner.

PR #215, opened 2020-04-09
Fix redirect to proper container

pgraverdy:
Nginx was redirecting front/bo calls to the same container.

### DIFF
--- a/.github/workflows/docker-build-run.yml
+++ b/.github/workflows/docker-build-run.yml
@@ -1,6 +1,6 @@
 name: Build Docker image and test container
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-build-run.yml
```diff
@@ -1,6 +1,6 @@
 name: Build Docker image and test container
 
-on: [push, pull_request]
+on: [pull_request]
```
rth, line 3: We still want this to build on master though? Not sure what's the best way to avoid redundant CI jobs.

rth: Yes, but sometimes CI can pass in PRs and fail on master. I think what we want is something like, https://help.github.com/en/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration

Let's merge as is, I'll open a separate PR.

PR #217, opened 2020-04-09
Trigger CI on PRs or push in master

rth:
Avoid redundant CI jobs when PRs are opened from upstream branches.

Makes CI trigger from either PRs to master or pushes to master (e.g. when the PR is merged).

### DIFF
--- a/.github/workflows/db-wipe-smoketest.yml
+++ b/.github/workflows/db-wipe-smoketest.yml
@@ -1,6 +1,14 @@
 name: DB wipe script smoke test
 
-on: [push, pull_request]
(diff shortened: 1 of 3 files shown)


pgraverdy:
@rth will this also work for draft pull requests ? (e.g., for a branch that work on docker, you may want to test each commit, so opening a draft PR would allow to launch these actions?)

Otherwise all fine for me.

rth:
>  will this also work for draft pull requests ?

I don't see why they wouldn't -- they are still PRs. If that's not the case when you try it, ping me.

PR #220, opened 2020-04-10
Add finess number + automate DB migration

olivierteboul:
Rewrote the migrator to discover discrepancy between the store and the actual db and do the necessary alterations.

to run the migration script:
python3 -m scripts.migrate_db

This should probably be run at every push

### DIFF
--- a/scripts/migrate_db.py
+++ b/scripts/migrate_db.py
@@ -1,30 +1,19 @@
 from absl import app
 from absl import flags
 from icubam import config
-from icubam.db import gsheets
 from icubam.db import migrator
-from icubam.db import sqlite
-from icubam.db import store
-from icubam.db import synchronizer
+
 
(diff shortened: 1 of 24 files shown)


scripts/migrate_db.py
```diff
@@ -1,31 +1,20 @@
...
   mgt.run()
 
-  # Now we add the region that was not in the old one.
-  sheets = gsheets.SheetsDB(cfg.TOKEN_LOC, cfg.SHEET_ID)
-  sync = synchronizer.StoreSynchronizer(sheets, mgt.new_db)
-  sync.sync_icus()
-
-
 
 if __name__ == "__main__":
```
dulacarnold, line 19: Add a click confirmation here, this is a non-trivial change to the DB file.

olivierteboul: ok

PR #222, opened 2020-04-10
Add periodic callback for predicu plots

rth:
This adds a `PeriodicCallback` to generate plots with `predicu` (external and for now private package) that are then rendered in the backoffice as part of https://github.com/icubam/icubam/pull/208. 

This works, assuming predicu is installed in the same environment as icubam. It has no effect if predicu is not installed.

However there are still open questions,
 - are we OK installing predicu on the production server in the same environment as icubam, assuming it becomes open-source. It's less well tested then icubam and has a few extra dependencies (matplotlib, seaborn, lxml). At worst if this callback fails for any reason, it shouldn't impact the rest of the backoffice server I imagine.
 - there is a risk of memory leaks with matplotlib. I can try to isolate this into a separate subprocess or thread, assuming that's compatible with async functions (not sure). 

I could look into it (and try to add more tests) if we decide to go this road, but I would like to have some feedback first. 

Other alternatives for deployment could be,
 - create a separate docker container, and either,
     - mount the DB there, install icubam and predicu and run a cron to generate figures periodically
     - or install only predicu, and download the data with an API token. This means that a token needs to be pre-generated in the DB when starting the BO server (if it's missing). This is potentially a bit more stressing for the server as it needs to generate a dump of the DB into a CSV.

Feedback welcome.


icubam/backoffice/server.py
```diff
@@ -80,37 +116,44 @@ def make_routes(self, path):
...
       route = os.path.join("/", self.root, r'static/extra-plots/(.*)')
-      self.routes.append(
-          (route, tornado.web.StaticFileHandler,
-          {'path': self.config.backoffice.extra_plots_dir})
-          )
+      self.routes.append((
+        route, tornado.web.StaticFileHandler, {
+          'path': self.config.backoffice.extra_plots_dir
+        }
+      ))
```
rth, line 123: Irrelevant formatting issues after running `yapf`. We should add it to CI.

rprimet: @rth OK I can get on it

olivierteboul:
Thanks Roman. Since there is some logics about predicu being installed or not, i would rather move all this into another module. Right now it seems like the most important part of the backoffice is this periodiccallback while it s really not.
Could you create a predicu_plots module and simply call predicu_plots.generate in the periodiccallback? This function might do nothing but logging a warning in case predicu is not installed.

rth:
> Could you create a predicu_plots module and simply call predicu_plots.generate in the periodiccallback? This function might do nothing but logging a warning in case predicu is not installed.

Will do. But otherwise you are OK with predicu running in the same environement as ICUBAM?

dulacarnold:
> Will do. But otherwise you are OK with predicu running in the same environment as ICUBAM?
I think it's OK to run the predicu code within the backoffice server, however it would be good to maybe package the timed callback into a seperate module.

rth:
OK I think this should be good to go.
 - the callback to generate plots is working, but disabled by default with `extra_plots_make_every=-1`
 - actually generating plots will currently fail because of typos in department issues https://github.com/icubam/icubam/issues/312 I would like to address that in a separate PR.
 - Similarly this adds a `predicu.data.normalize_colum_names` to convert column names to those expected by plots. The right way to fix this is instead go through each plot and change column names / dtype to those used in the DB.
 - This also adds some consistency checks between DB schema and the fake data we currently use to test plots.

Overall there is still work on predicu integration and https://github.com/icubam/icubam/issues/301 is fundamental for detecting such issues before deployment.

icubam/predicu/test_utils.py
```diff
@@ -13,6 +13,8 @@ def load_test_data():
   test_bc = pd.read_csv(test_bc_path, index_col=False, comment="#")
   test_bc['date'] = pd.to_datetime(test_bc['date']).dt.date
   test_bc['datetime'] = pd.to_datetime(test_bc['datetime'])
+  # these should be string not int
+  test_bc['icu_name'] = test_bc['icu_name'].astype(object)
```
dulacarnold, line 17: Should we then just cast to str here?

icubam/predicu/analytics_server.py
```diff
@@ -0,0 +1,54 @@
+from pathlib import Path
+
+from absl import logging  # noqa: F401
+
+from .plot import generate_plots
```
dulacarnold, line 5: Do we want to use relative imports vs icubam.predicu.plot?  I'm always weary of python relative imports (but probably because I don't understand/use them well).

icubam/backoffice/server.py
```diff
@@ -9,21 +9,12 @@
...
-from icubam.backoffice.handlers import icus
-from icubam.backoffice.handlers import login
-from icubam.backoffice.handlers import logout
-from icubam.backoffice.handlers import maps
-from icubam.backoffice.handlers import messages
-from icubam.backoffice.handlers import operational_dashboard
-from icubam.backoffice.handlers import regions
-from icubam.backoffice.handlers import tokens
-from icubam.backoffice.handlers import upload
-from icubam.backoffice.handlers import users
```
rth, line 26: Not really sure what happened here, but I didn't do it manually. maybe yapf?

icubam/predicu/analytics_server.py
rth: Implicit relative imports were problematic in Python 2 e.g. `import plot`. In python 3 explicit relative imports are fine e.g. `from . import plot` or `from .plot import generate_plots`. This is done with respect to the current directory and they are non ambiguous: https://mail.python.org/pipermail/python-dev/2010-October/104476.html

But reverted to absolute imports here for now.

PR #224, opened 2020-04-10
run yapf in CI, run yapf on whole codebase first to be conformant

rprimet:
Warning giant diff incoming.

Following a suggestion by @rth that we run yapf on CI, here is a candidate workflow.
I had to run yapf on the whole codebase first which explains this large diff 

Another possibility would be to add a pre-commit hook (pre-commit.com or otherwise) to avoid pushing non-yapfing code in the first place but that might be a second PR?

### DIFF
--- a/.github/workflows/yapf.yml
+++ b/.github/workflows/yapf.yml
@@ -1,0 +1,24 @@
+name: run yapf on source code
+
+on: push
+
+jobs:
+  yapf:
+    runs-on: ubuntu-latest
+    steps:
+    # https://github.com/marketplace/actions/setup-miniconda
+    - uses: actions/checkout@v2
+    - uses: goanpeca/setup-miniconda@v1
+      with:
+         auto-update-conda: true
+         activate-environment: anaconda-client-env
+         python-version: 3.8
+         auto-activate-base: false
+    - name: Install yapf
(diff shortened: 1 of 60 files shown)


.github/workflows/yapf.yml
```diff
@@ -0,0 +1,24 @@
...
+         python-version: 3.8
+         auto-activate-base: false
+    - name: Install yapf
+      shell: bash -l {0}
+      run: |
+        pip install yapf
+    - name: run yapf
+      shell: bash -l {0}
+      run: |
+        yapf -r --diff .
```
rth, line 24: Does it exit with a non zero status if there is some diff?

rprimet: Yes I believe it does, the `--diff` option does that and this is what the manual suggests in ci

```
(base) (icubam)romain@plume:~/inria/icubam$ cat foo.py
def foo():
        print("hi")
(base) (icubam)romain@plume:~/inria/icubam$ yapf -r --diff . && echo Previous command terminated cleanly
--- ./foo.py	(original)
+++ ./foo.py	(reformatted)
@@ -1,2 +1,2 @@
 def foo():
-        print("hi")
+  print("hi")

```

PR #230, opened 2020-04-10
Consent page

olivierteboul:
This PR changes many things:
- send the user_id and name in the update form
- as a consequence the links are now changed, to contained the encoded user_id
- store the user in the baseHandler
- the user is authenticated if it has the cookie and it has not consented yet or has answered yes
- changes also impact the way we write the bedcount (now with the user, old todo is gone)
- Added a modal in the html
- this modal only launches if there is some existing consent html file specified in the config and the user has not answered yet.


icubam/www/templates/update_form.html
```diff
@@ -3,10 +3,40 @@
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <h4 class="modal-title">{{_("Consent")}}</h4>
+      </div>
+      <div class="modal-body">
+        <p>{% raw consent %}</p>
+      </div>
+      <div class="modal-footer">
+        <input type="checkbox" id="double_check" title="{{_('I understand what I am doing.')}}" name="double_check"/>
+        <label for="double_check">{{_('I understand what I am doing.')}}</label>
```
VincentRouvreau, line 24: Here, you have to be more specific cf. https://www.cnil.fr/fr/RGPD-exemple-information-acces-aux-locaux-professionnels-par-badge :
ICUBAM's goal is to provide real-time information on intensive care unit (ICU) bed availability in hospitals. The following personal data from doctors working inside ICU are stored:
* First and last name
* email
* cell phone number
* job title, hospital, area, department and city
The data will be only available by the administrators of the service for maintenance purpose. The data will be stored as long as the COVID-19 pandemy will last. If you do not accept our conditions, your user personal data will be erased from the database within a month. You can asked for your personal data stored or you ask to be deleted from the user simply by sending us an email at the address contact@icubam.net

Maybe a french translation would be good too.

olivierteboul: Thanks Vincent. I am not putting here the text of the RGPD document. The main reason is two fold. 1) the lawyers should write this document. 2) this is a document about exploiting ICUBAM not about the software itself. Someone in another country may want to use icubam and the text that would need to be used would be different.

So this PR's goal is to give the possibility for whoever decides to operate icubam in france or elsewhere to do it by simply adding the file path to a config and not to mix up the icubam open source software with its exploitation by some institution or company.

In practice, with this PR, in the config, if there is a field `server.consent` that points to a valid path to some html / text file this file would be read and pasted in the modal. Otherwise there will be no modal shown and no consent asked.

PR #233, opened 2020-04-10
BedCount Ingestion with Synchronizer

dulacarnold:


icubam/db/synchronizer.py
```diff
@@ -8,6 +8,11 @@
 
 ICU_COLUMNS = ['name', 'region', 'dept', 'city', 'lat', 'long', 'telephone']
 USER_COLUMNS = ['icu_name', 'name', 'telephone', 'description']
+BC_COLUMNS = [
```
olivierteboul, line 11: btw you can get those with `store.BedCount.get_column_names(include_relationship=False)`
Same above

icubam/db/synchronizer.py
```diff
@@ -113,6 +118,26 @@ def sync_users(self, users_df, force_update=False):
...
+    # Make sure each bedcount has an existent ICU:
+    icu_diff = icu_names - set(db_icus.keys())
+    if icu_diff is not None:
+      KeyError("Missing ICUs in DB: {icu_diff}. Please add them first.")
+
+    # Now we are sure all ICUs are present so we can insert without checking:
+    for bc in bedcounts_df.iterrows():
+      item = None
+      self.db.update_bed_count_for_icu(
+        None, store.BedCount(**item), force=True
```
olivierteboul, line 138: I believe instead you could also pass a user to this (these) function. Since this might be used from the backoffice where there is user calling this function. Hence if would do
`self.db.update_bed_count_for_icu(user, store.BedCount(**item), force=user is None)`

dulacarnold: I added something that should work, let me know if it's good (once I push).

icubam/db/synchronizer.py
dulacarnold: So that is the full set, but since I use this to test that the provide dataframe has the correct columns, I want only a subset (the necessary columns).... So either I use the ones from the store and explicitly delete unnecessary columns, let's discuss offline.

icubam/db/test_synchronizer.py
```diff
@@ -5,6 +5,44 @@
...
+      bc = {'icu_name': icu_name,
+        'n_covid_occ': value,
+        'n_covid_free': value,
+        'n_ncovid_occ': value,
+        'n_ncovid_free': value,
+        'n_covid_deaths': value,
+        'n_covid_healed': value,
+        'n_covid_refused': value,
+        'n_covid_transfered', value,
+        'timestamp'
```
tgy, line 30: this does not compile

PR #236, opened 2020-04-12
Refactor config paths

maximedenes:
This PR follows suggestions made in https://github.com/icubam/icubam/pull/117#discussion_r401226576

As a side effect, it also makes instructions in `doc/install.md` succeed when they were previously failing to launch `run_server.py` because of a missing path to the `.env` file.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -15,7 +15,7 @@ Steps:
 
 ### Configuration
 
-Create a `resources/.env` file containing the following keys:
+Create a `resources/icubam.env` file containing the following keys:
 ```
 SHEET_ID=
 TOKEN_LOC=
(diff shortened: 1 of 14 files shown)


dulacarnold:
This looks good to me, just want to check with @olivierteboul before LGTM.

maximedenes:
@dulacarnold just a small note: for this PR, it would have been better to not squash the commits IMHO. One was fixing making the default values shared, the other one changing the actual value. They were really two different changes.

dulacarnold:
My bad, I'm still getting a handle on Git's merge scheme. I'm assuming a rebase + merge would have been preferable. To save me from making similar mistakes, don't hesitate to specify in the description!

April 14 2020 12:21 AM, "Maxime Dénès"  wrote:
	@dulacarnold (https://github.com/dulacarnold) just a small note: for this PR, it would have been better to not squash the commits IMHO. One was fixing making the default values shared, the other one changing the actual value. They were really two different changes.

PR #239, opened 2020-04-12
Add the possibility to restrict an API KEY to one or several regions (MAP only)

olivierteboul:
- if no regions are specified, it gives access to the whole map
- Otherwise restricts the map to some regions.
- no support yet for db access


icubam/map_builder.py
```diff
@@ -43,10 +43,18 @@ def to_map_data(self, tree, level):
...
+    center_icu: Optional[store.ICU] = None,
+    regions: Optional[List[int]] = None,
+    level: str = 'dept'
+  ):
     tree = icu_tree.ICUTree()
-    tree.add_many(self.db.get_icus(), self.db.get_latest_bed_counts())
+    icus = self.db.get_icus()
+    if regions is not None and len(regions) > 0:
+      icus = [icu for icu in icus if icu.region_id in regions]
+    tree.add_many(icus, self.db.get_latest_bed_counts())
```
rth, line 57: I there are no regions, it would use ICUs from all regions. Is that desirable? I would have expected the above condition to only be `regions is not None` same as in the other file.

olivierteboul: indeed this is a bit of the same here, since regions in the store comes as a list (maybe empty), so we cannot set from the backoffice or the store the regions to be None. On the backoffice we choose not to have region restrictions and hence get an empty list.

PR #240, opened 2020-04-12
[predicu] refactor data caching

tgy:


icubam/predicu/data/icu_name_to_department.json
```diff
@@ -1 +1 @@
...
... "Ille-et-vilaine", "SC-CLIN-NAVARRE": "Pyr\u00e9nn\u00e9es Atlantique", "HTP-Ped": "Bas-Rhin", "Dreux": "Eure-et-Loir", "St-Avold": "Moselle", "Bourges": "Cher", "Tours-R-NChir": "Indre-et-Loire", "Avranches": "Manche", "BDX-PELL-SC-3": "Gironde", "Annonay-CH": "Loire", "BDX-REA-GHS": "Gironde", "Brest-RC": "Finistere", "Brest-MIR": "Finistere", "Saint-Et-Hop-Prive-Loire": "Loire", "C-Rhena": "Bas-Rhin", "Nancy-Ped": "Meurthe-et-Moselle", "Quimper": "Finistere", "Troyes": "Aube", "NHC-ChirC": "Bas-Rhin", "CHR-Mercy": "Moselle", "Grenoble-AR": "Isere", "Bicetre": 93, "A. Beclere": 92, "A. Pare": 93, "Avicenne": 93, "Beaujon": 93, "Bayonne": "Pyr\u00e9nn\u00e9es Atlantique", "Bayonne USC": "Pyr\u00e9nn\u00e9es Atlantique", "BDX R PICQUE": "Gironde", "Alencon": "Orne", "Libourne USC": "Gironde", "Mont de Marsan USC": "Landes", "BDX St AUG": "Gironde", "Perigueux": "Dordogne", "Dax USC": "Landes", "Nantes-MIR": "Loire-Atlantique", "Agen": "Lot et Garonne", "Flers": "Orne", "BDX MIR HSA USC": "Gironde", "Niort": "Deux-Sevre", "a": "Marne", "c": "Marne", "b": "Marne", "d": "Bas-Rhin", "e": "Haut-Rhin", "Agen USC": "Lot et Garonne", "Angers-MIR": "Maine-et-Loire", "Lyon-MIR-CR": "Rhone"}
```
rth, line 1: Why does this file keep changing? Is it generated by one of the load functions? Should we exclude it from git and let it be re-generated locally as needed?

icubam/predicu/plot/__init__.py
```diff
@@ -103,74 +113,60 @@ def plot_int(
...
 
 
-def plot(plot_name, data, **plot_args):
+def plot(
+  plot_name: str,
+  cached_data: Dict[str, pd.DataFrame],
+  output_dir: str = DEFAULTS["output_dir"],
+  output_type: str = DEFAULTS["output_type"],
+  api_key: str = DEFAULTS["api_key"],
+  matplotlib_style: str = DEFAULTS["matplotlib_style"],
```
rth, line 122: The default are a bit convoluted. If you change the `DEFAULTS` dict globally and then use this functions, the parameter defaults will likely not change.

Maybe instead set default to None, and then,
```py
if api_key is None:
    api_key = DEFAULTS["api_key"]
```
or even better just remove the `DEFAULTS` dict and write them plainly in these 2 functions. Most are None anyway.

icubam/predicu/plot/__init__.py
```diff
@@ -103,74 +113,60 @@ def plot_int(
...
+):
   plot_module = __import__(f"{plot_name}", globals(), locals(), ["plot"], 1)
```
rth, line 124: In the long term, I think we should rather manually import all plot modules at the top of `__init__`, e.g.
```oy
from . import barplot_deaths_per_day
```
and then use something like,
```py
plot_module = globals()[plot_name]
```
inside this function. Import of a user define path is bad, as it's not far from arbitrary code execution as long as one is able to put some file in that folder.

icubam/predicu/plot/__init__.py
```diff
@@ -103,74 +113,60 @@ def plot_int(
...
       filepath=output_path,
       figure=fig,
       **tikzplotlib_kwargs,
       standalone=True,
     )
   elif output_type in ["png", "pdf"]:
-    fig.savefig(
-      os.path.join(plot_args["output_dir"], f"{plot_name}.{output_type}")
-    )
+    fig.savefig(os.path.join(output_dir, f"{plot_name}.{output_type}"))
```
rth, line 142: For future refacorning, let's use, [pathlib](https://docs.python.org/3/library/pathlib.html) instead of os.path. Here it would have been,
```py
from pathlib import Path

def ...
   ...
   output_dir = Path(output_dir)
   ...
   fig.savefig(str(output_dir / f"{plot_name}.{output_type}")))
```

icubam/predicu/tests/data/fake_all_bedcounts_2020-04-08_16h41.csv
```diff
@@ -1,101 +0,0 @@
-icu_name,date,datetime,icu_dept,n_covid_deaths,n_covid_healed,n_covid_transfered,n_covid_refused,n_covid_free,n_ncovid_free,n_covid_occ,n_ncovid_occ
```
rth, line 1: I would rather we re-generated this from the DB in memory in each test session, and made sure *.csv is in .gitignore to avoid commiting production dumps. Same comment for the .h5 below.

Otherwise we won't be sure that testing on this file is consistent with testing on production data.

We can do that, can't we?

icubam/predicu/data/icu_name_to_department.json
olivierteboul: i don t even get why we do not get those from the database to begin with...

tgy: we do, and actually this file is updated from the database. this is due to some pre-ICUBAM data preprocessing but @dulacarnold and I are currently working on integrating this data to ICUBAM and we won't have to use this file afterwards... i know this looks weird but it was necessary. i will probably remove it in this PR

tgy: today, @dulacarnold and i pushed the pre-ICUBAM data to the prod's database. i was able to remove this file.

icubam/predicu/plot/__init__.py
tgy: done

icubam/predicu/plot/__init__.py
tgy: agreed. let's fix that in another PR?

icubam/predicu/tests/data/fake_all_bedcounts_2020-04-08_16h41.csv
tgy: re-generate from which DB in memory? not sure i understand

icubam/predicu/plot/__init__.py
rth: Here, is fine, I'm not asking for changes in this PR.  I just mean that scripts that use a lot of paths manipulation are significantly better off using pathlib than os.path https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/ It helps readability for one thing.

icubam/predicu/data.py
```diff
@@ -4,23 +4,55 @@
...
 import pandas as pd
 from lxml import html
 
 BASE_PATH = os.path.dirname(__file__)
 
+DATA_SOURCES = {"icubam", "bedcounts", "public", "combined_bedcounts_public"}
+
 DATA_PATHS = {
-  "icu_name_to_department": "data/icu_name_to_department.json",
-  "pre_icubam": "data/pre_icubam_data.csv",
```
rth, line 16: Great!

icubam/predicu/data.py
```diff
@@ -44,58 +76,58 @@
...
+  d = load_if_not_cached(
+    "icubam", cached_data, api_key=api_key, icubam_host=icubam_host
+  )
   if clean:
     d = clean_data(d, spread_cum_jump_correction)
   d = d.sort_values(by=["date", "icu_name"])
   if max_date is not None:
     logging.info("data loaded's max date will be %s (excluded)" % max_date)
     d = d.loc[d.date < pd.to_datetime(max_date).date()]
+  d = d.loc[d.department.isin(DEPARTMENTS_GRAND_EST)]
```
rth, line 109: I find this strange. Nothing in the function name or description suggests that discards data for departments other than Grand Est.  Are sure about it?

I see it was done in plots earlier. If you want to avoid repetition, maybe, pass an optional `deparment=None` (i.e. no filtering by default) to `load_bedcounts` and then use load_bedcounts(department="Grand-Est") in all plots. The danger of current factorization is that from plots it's very hard to know this filtering is happening.

icubam/predicu/data.py
```diff
@@ -4,23 +4,55 @@
...
+  )
+
+
+def load_france_departments():
+  return pd.read_json(DATA_PATHS["departments"])
+
+
+DEPARTMENTS_GRAND_EST = [
+  "Ardennes", "Aube", "Marne", "Haute-Marne", "Meurthe-et-Moselle", "Meuse",
+  "Moselle", "Bas-Rhin", "Haut-Rhin", "Vosges"
```
rth, line 39: So in the mid-term all this will be loaded from the DB, right? Maybe let's add an inline comment about it?

icubam/predicu/plot/__init__.py
tgy: Very convincing article. I'll use that from now on :)

icubam/predicu/tests/data/bedcounts.csv
```diff
@@ -0,0 +1,1054 @@
+icu_name,date,datetime,department,n_covid_deaths,n_covid_healed,n_covid_transfered,n_covid_refused,n_covid_free,n_ncovid_free,n_covid_occ,n_ncovid_occ
```
rth, line 1: I am very scared by these .csv files, because just by looking at it it's hard to know that this is not senstivite production data. If necessary, please add a comment on the top to indicate that we would want to remove it in the future in favor of using DB, and rename them to clearly indicate that it's only test data;

rth: Also let's add a `*.csv` to `.gitignore` to make it harder to commit sensitive csv by accident.

icubam/predicu/data.py
tgy: the original analysis was on the Grand Est region and any ICU that was not in the Grand Est region was discarded from the analysis. this should be removed, i agree.

icubam/predicu/tests/data/bedcounts.csv
tgy: i believe them being located in `icubam/predicu/tests/data` makes it pretty clear already that this is test data. this data was obfuscated by way of randomisation. it contains random numbers.

icubam/predicu/data.py
tgy: yes i think we can now use some `icu_region` attribute of `BedCounts` objects no? if so, this is useless and can be removed. good point.

icubam/predicu/tests/data/bedcounts.csv
rth: > i believe them being located in icubam/predicu/tests/data makes it pretty clear already that this is test data.

Great! To me it wasn't obvious. If it contains random numbers, then let's add this with a comment on the top,
```
# This file contains a randomly generated dataset for testing
```

icubam/predicu/tests/data/fake_all_bedcounts_2020-04-08_16h41.csv
rth: Well to run ICUBAM tests, one would run `scripts/populate_db_fake.py` https://github.com/icubam/icubam/blob/master/doc/install.md#pre-populate-db-with-test-data so we should be able to use that same data for our tests, without needing to rely on pre-generated CSV data.

That fake data generation script might need to be updated though. Not asking to do it here.

icubam/predicu/tests/data/bedcounts.csv
tgy: pandas did not accept this comment

rth: Need to pass `comment="#"` to pd.read_csv..

tgy: thanks, done.

PR #249, opened 2020-04-13
Lint shell scripts as part of CI

marccarre:
### Why?

To help find bugs early and enforce standards & best practices.
(Since `yapf` was added recently, I thought this would be one more step in that direction. Feel free to close if undesirable for whatever reason, though. 🙂)

### What?

- Lint shell scripts as part of CI.
- Fix all warnings raised by `shellcheck`:

  ```console
  find . -type f -name '*.sh' | xargs spellcheck
  ```

    - SC2086: https://github.com/koalaman/shellcheck/wiki/SC2086
    - SC2215: https://github.com/koalaman/shellcheck/wiki/SC2215
    - SC2162: https://github.com/koalaman/shellcheck/wiki/SC2162
    - SC2128: https://github.com/koalaman/shellcheck/wiki/SC2128

### DIFF
--- a/docker/scripts/init-letsencrypt.sh
+++ b/docker/scripts/init-letsencrypt.sh
@@ -12,7 +12,7 @@ email="email@example.org" # Adding a valid address is strongly recommended
 staging=0 # Set to 1 if you're testing your setup to avoid hitting request limits
 
 if [ -d "$data_path" ]; then
-  read -p "Existing data found for $domains. Continue and replace existing certificate? (y/N) " decision
+  read -r -p "Existing data found for ${domains[0]}. Continue and replace existing certificate? (y/N) " decision
   if [ "$decision" != "Y" ] && [ "$decision" != "y" ]; then
     exit
   fi
@@ -27,9 +27,9 @@ if [ ! -e "$data_path/conf/options-ssl-nginx.conf" ] || [ ! -e "$data_path/conf/
   echo
 fi
 
-echo "### Creating dummy certificate for $domains ..."
-path="/etc/letsencrypt/live/$domains"
-mkdir -p "$data_path/conf/live/$domains"
+echo "### Creating dummy certificate for ${domains[0]} ..."
+path="/etc/letsencrypt/live/${domains[0]}"
+mkdir -p "$data_path/conf/live/${domains[0]}"
 docker-compose -f docker/docker-compose.yml --project-directory . \
   run --rm --entrypoint "\
   openssl req -x509 -nodes -newkey rsa:1024 -days 1\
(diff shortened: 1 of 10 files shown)


dulacarnold:
Will wait on @pgraverdy 's approval but +1 to this.

marccarre:
If we proceed with this PR, and we're keen with the idea, I may add a commit hook like for `yapf`. (Will do that separately/in another PR.)

tgy:
thanks, maybe also add a `isort` check?

marccarre:
> thanks, maybe also add a `isort` check?

[`isort`](https://github.com/timothycrosley/isort) sgtm 👍, but I'd rather do that in another PR if that's fine with you @tgy.

docker/scripts/init-letsencrypt.sh
```diff
@@ -27,9 +27,9 @@ if [ ! -e "$data_path/conf/options-ssl-nginx.conf" ] || [ ! -e "$data_path/conf/
   echo
 fi
 
-echo "### Creating dummy certificate for $domains ..."
-path="/etc/letsencrypt/live/$domains"
-mkdir -p "$data_path/conf/live/$domains"
+echo "### Creating dummy certificate for ${domains[0]} ..."
```
pgraverdy, line 30: @marccarre 
In the original script, the "domains" variable may contain multiple hostnames (e.g., domains=(example.org www.example.org)).

The linting changes the code so that only the first item is used (I agree that will be the usual case) but then will fail for this special case. 

Is there a way to parametrize/disable this change, or should we change the initial script (in particular the initialisation of the variable ?

pgraverdy: Otherwise, I'm fine with this PR. Maybe we can put a note/caveat in the doc and open an issue for later ?

marccarre: I may be missing something 🤔, but it looks like the linter correctly raised [`SC2128`](https://github.com/koalaman/shellcheck/wiki/SC2128) here? Indeed, using `$domains` only prints the first value of the array, so the script may have had bugs from the very beginning?

```console
$ cat test_array.sh
#!/bin/bash
set -x
domains=(example.org www.example.org)
echo $domains
echo "$domains"
echo "${domains[0]}"
echo "${domains[@]}"

$ ./test_array.sh
+ domains=(example.org www.example.org)
+ echo example.org
example.org
+ echo example.org
example.org
+ echo example.org
example.org
+ echo example.org www.example.org
example.org www.example.org
```

So, I guess it is a matter of where we want to use the full array, and where we may have bugs and want to iterate over the array?
For example:

1. We may want:
    ```console
    echo "### Creating dummy certificate for ${domains[@]} ..."
    ```
    instead of:
    ```console
    echo "### Creating dummy certificate for $domains ..."
    ```
    in order to print all values?

2. But we may want a `for` loop in places like:
    ```console
    docker-compose -f docker/docker-compose.yml --project-directory . \
      run --rm --entrypoint "\
      rm -Rf /etc/letsencrypt/live/$domains && \
      rm -Rf /etc/letsencrypt/archive/$domains && \
      rm -Rf /etc/letsencrypt/renewal/$domains.conf" certbot
    ```
    (Or maybe we'd want to revisit the script more broadly?)

Am I missing something?
WDYT @pgraverdy?
(If my analysis is correct, happy to revisit this PR to have `for` loops where relevant.)

rth:
> Maybe we can put a note/caveat in the doc and open an issue for later ?

Yes, please create a `doc/contributing.md` add a sentence about linting bash scripts there, and link it from `doc/README.md`. Thanks!

PR #253, opened 2020-04-13
Fix #251: /list_tokens HTTP 500

### LINKED ISSUE #251: backoffice (v0.3.0): /list_tokens HTTP 500 internal server error
Getting a HTTP 500 internal server error on /list_tokens in backoffice with fake DB data.

- on master v0.3.0 (31fffff)
- generated the fake db before launching the Docker container.
- launched in dev mode locally using `docker-compose -f docker/docker-compose-core.yml --project-directory .  up
`
- URL http://localhost:8890/bo/ then http://localhost:8890/bo/list_tokens

```
icubam-bo-server  | HTTPServerRequest(protocol='http', host='localhost:8890', method='GET', uri='/bo/list_tokens', version='HTTP/1.1', remote_ip='172.24.0.1')
icubam-bo-server  | Traceback (most recent call last):
icubam-bo-server  |   File "/root/miniconda3/envs/icubam/lib/python3.8/site-packages/tornado/web.py", line 1697, in _execute
icubam-bo-server  |     result = method(*self.path_args, **self.path_kwargs)
icubam-bo-server  |   File "/root/miniconda3/envs/icubam/lib/python3.8/site-packages/tornado/web.py", line 3174, in wrapper
icubam-bo-server  |     return method(self, *args, **kwargs)
icubam-bo-server  |   File "/home/icubam/icubam/backoffice/handlers/tokens.py", line 46, in get
icubam-bo-server  |     self.render_list(
icubam-bo-server  |   File "/home/icubam/icubam/backoffice/handlers/base.py", line 38, in render_list
icubam-bo-server  |     columns = json.dumps([x['key'] for x in data[0]])
icubam-bo-server  | IndexError: list index out of range
```

maximedenes:
IIUC, https://github.com/icubam/icubam/commit/fd464da6504be58302fa8d6132802088bacac807 is trying to extract column names from the first item in data passed to `render_list`, which fails if the data is empty.

I believe this fix should restore previous behavior on empty lists.

I haven't explored the testing infrastructure yet, so I'm not sure how to add a regression test. Any guidance welcome.

Fixes #251.

### DIFF
--- a/icubam/backoffice/handlers/base.py
+++ b/icubam/backoffice/handlers/base.py
@@ -35,7 +35,8 @@ class BaseHandler(tornado.web.RequestHandler):
 
   def render_list(self, data, objtype, create_handler=None, **kwargs):
     route = None if create_handler is None else create_handler.ROUTE
-    columns = json.dumps([x['key'] for x in data[0]])
+    item = data[0] if data else []
+    columns = json.dumps([x['key'] for x in item])
     return self.render(
       "list.html",
       data=data,


icubam/backoffice/handlers/base.py
```diff
@@ -35,7 +35,7 @@ def render(self, path, **kwargs):
 
   def render_list(self, data, objtype, create_handler=None, **kwargs):
     route = None if create_handler is None else create_handler.ROUTE
-    columns = json.dumps([x['key'] for x in data[0]])
+    columns = [] if not data else json.dumps([x['key'] for x in data[0]])
```
rth, line 38: The output of json.dumps is a string, so we would rather want something like,

```suggestion
    if not data:
        columns = '[]'
    else:
        columns = json.dumps([x['key'] for x in data[0]])
```

to preserve type. 

For tests, we would likely need to add a test in `backoffice/server_test.py` that tries to render the `/list-tokens` endpoint when there are none but I haven't looked in detail. See https://github.com/icubam/icubam/issues/251#issuecomment-612951088 for related discussion.

maximedenes: Good catch, thanks! I'm too used to statically typed languages, where you can rely on the compiler :)

olivierteboul: good catch. I would rather do something like
```
item = data[0] if data else []
columns = json.dumps([x['key'] for x in item])
```

maximedenes:
@olivierteboul Thanks, I followed your suggestion, and confirmed this is detected by existing tests once #256 is merged. So I believe this is ready for merge.

rth:
> I'm too used to statically typed languages, where you can rely on the compiler :)

[Mypy](https://github.com/python/mypy) can help some.

Issue #197, opened 2020-04-08
Code coverage of handlers


rth:
Related to https://github.com/icubam/icubam/issues/109, currently www handlers are shown as mostly not being covered by tests,
https://codecov.io/gh/icubam/icubam/tree/520140e5d480e8f2605544530c8ff42e08461294/icubam/www/handlers

aside for the version and static endpoints,

![Screenshot_2020-04-08 Code coverage done right ](https://user-images.githubusercontent.com/630936/78786069-76a1c000-79a8-11ea-80cb-0e31f708025c.png)

We should add tests to at least render the corresponding pages.

The same applies to BO  https://codecov.io/gh/icubam/icubam/tree/520140e5d480e8f2605544530c8ff42e08461294/icubam/backoffice/handlers

though there it's strange since https://github.com/icubam/icubam/blob/520140e5d480e8f2605544530c8ff42e08461294/icubam/backoffice/server_test.py#L50 should have tested them. I suspect this test is not working as expected, which would also explain some strange behavior I was seeing when testing the `/operational-dasboard` endpoint https://github.com/icubam/icubam/pull/159#issuecomment-609042864

Either that or tornado do not allow to compute code coverage properly for requests originating in tests.

fpquintao:
The test coverage for the BO handlers is super low for sure.

maximedenes:
The tests indeed do not work properly. It seems related to authentication, because exposing the endpoints without authentication seems to make the test actually run.

My bet is that there are two problems:
- the bo handlers test does not pass authentication credentials correctly
- the authentication system returns a 302 in order to redirect to the login page, probably triggering a redirection in the test which then reaches a 200 on the login page and sees it as a success

I'll try to fix the second problem, then the first.

rth:
> the authentication system returns a 302 in order to redirect to the login page, probably triggering a redirection in the test which then reaches a 200 on the login page and sees it as a success

Ahh, that would explain it! Thanks for investigating.

rth:
Maybe we should check that the final response URL is the same as the request URL.

maximedenes:
> Maybe we should check that the final response URL is the same as the request URL.

It would probably work, but I have a slightly more robust patch, which follows redirections only for tests where it is expected.

So I confirmed the second point above, and fixed it. Now working on the first (fixing authentication).

It could be a good idea to wait that I fix the tests before merging other backoffice-related PRs, since the tests are actually not effective.

maximedenes:
[...]

PR #256, opened 2020-04-13
Fix endpoints tests

maximedenes:
This PR:
- Makes backoffice tests detect when authentication fails for an endpoint. Previously, it was considered as a success because of a redirection to the login page.
- Fixes authentication from tests by using Tornado's `current_user` instead of rolling our own `user` field.

The second part especially needs careful review. We may want to do the same for `www`, but let's see how things work out here first.

TODO:
- [x] Add a comment about `current_user` being assigned the result of `get_current_user` by Tornado
- [x] In templates, use Tornado's user variable directly

DO NOT SQUASH this PR, the history is clean and meaningful.

### DIFF
--- a/icubam/backoffice/handlers/base.py
+++ b/icubam/backoffice/handlers/base.py
@@ -26,7 +26,6 @@ class BaseHandler(tornado.web.RequestHandler):
     status = self.application.server_status
     super().render(
       path,
-      this_user=self.current_user,
       root=self.root_path,
       server_status=status,
       **kwargs
--- a/icubam/backoffice/handlers/base.py
+++ b/icubam/backoffice/handlers/base.py
@@ -19,7 +19,6 @@ class BaseHandler(tornado.web.RequestHandler):
       self.root_path = '/{}/'.format(root)
     else:
       self.root_path = '/'
-    self.user = None
 
   def render(self, path, **kwargs):
     # This dictionary is updated by a PeriodicCallback in the
@@ -27,7 +26,7 @@ class BaseHandler(tornado.web.RequestHandler):
     status = self.application.server_status
     super().render(
       path,
-      this_user=self.user,
+      this_user=self.current_user,
       root=self.root_path,
       server_status=status,
       **kwargs
@@ -49,16 +48,14 @@ class BaseHandler(tornado.web.RequestHandler):
   def get_template_path(self):
     return os.path.join(self.PATH, 'templates/')
 
+  # Tornado's @tornado.web.authenticated decorator will put the result of this
+  # function in the `current_user` field of the handler
+  # See https://www.tornadoweb.org/en/stable/guide/security.html#user-authentication
   def get_current_user(self):
-    if self.user is not None:
-      return self.user
-
     userid = self.get_secure_cookie(self.COOKIE)
     if not userid:
       return None
-
-    self.user = self.db.get_user(int(tornado.escape.json_decode(userid)))
-    return self.user
+    return self.db.get_user(int(tornado.escape.json_decode(userid)))
 
   def get_user_locale(self):
     locale = self.get_query_argument('hl', default=self.config.default_locale)
--- a/icubam/backoffice/handlers/bedcounts.py
+++ b/icubam/backoffice/handlers/bedcounts.py
@@ -45,7 +45,7 @@ class ListBedCountsHandler(base.BaseHandler):
   @tornado.web.authenticated
   def get(self):
     locale = self.get_user_locale()
-    icus = self.db.get_managed_icus(self.user.user_id)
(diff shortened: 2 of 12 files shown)


icubam/backoffice/handlers/base.py
```diff
@@ -19,15 +19,14 @@ def initialize(self):
...
-    self.user = None
 
   def render(self, path, **kwargs):
     # This dictionary is updated by a PeriodicCallback in the
     # BackofficeApplication
     status = self.application.server_status
     super().render(
       path,
-      this_user=self.user,
+      this_user=self.current_user,
```
fpquintao, line 29: perhaps I'm missing something, but self.current_user is never set.

maximedenes: It is set by the `@tornado.web.authenticated` decorator.

maximedenes:
The tests are failing because they now reveal #251. We should probably merge #253 first.

icubam/backoffice/handlers/base.py
fpquintao: Ahhhh that's great, I'm double checking the documentation.
Perhaps one little comment about it in the get_current_user method?

maximedenes: See https://www.tornadoweb.org/en/stable/guide/security.html#user-authentication

maximedenes: Sure, I'll add a comment.

fpquintao: I saw it later. Apparently we don't even need to inject the self.current_user into the template as this_user since the user is already injected according to the docs.

But this can be fixed in a follow-up CL (i.e., cleaning up the templates in favour of current_user instead of this_user). Can you add a TODO for that?

Thanks a lot for the nice fix.

maximedenes: Ah ah, I also realized while sending you the pointer :)

I can do it in this PR, it is morally part of the same change.

olivierteboul: cool ! Good job ! Feel free to do it either in this or in a subsequent PR.

rth:
Thanks you @maximedenes ! This will certainly address most of #197

Could you sync with master to see if CI now passes with /list-tokens fix? For the /operational-dashboard I'll look into it in a separate PR.

maximedenes:
> Could you sync with master to see if CI now passes with /list-tokens fix?

Yes, I'll address the remaining points and rebase to see what CI says.

maximedenes:
This PR seems to reveal another problematic test : ListMessagesHandler. Is the message server started in backoffice tests?

rth:
I would say comment everything that fails (which wasn't running on master anyway), and we'll fit it separate. This PR has already a fair amount of changes.

maximedenes:
> I would say comment everything that fails (which wasn't running on master anyway), and we'll fit it separate. This PR has already a fair amount of changes.

Sounds good.

maximedenes:
This PR should be ready modulo CI. To anyone merging, please DO NOT SQUASH this PR. I made sure the history is clean and meaningful.

rth:
Wonderful, thank you @maximedenes !

(One limitation of rebase & merge is that it doesn't link to the PR directly in the commit message and one has to click on the commit to discover which PR it belongs to).

rth:
Yapf check is not running on PRs, or am I missing something? Because the CI was green here, and yapf check failed on master.

maximedenes:
> One limitation of rebase & merge is that it doesn't link to the PR directly in the commit message and one has to click on the commit to discover which PR it belongs to

True, but on the other hand, small and separate commits have high value for later understanding of the development, and they can't be recovered once lost. On the other hand, I wouldn't be surprised that GitHub improves the handling of such PRs.

maximedenes:
> Yapf check is not running on PRs, or am I missing something? Because the CI was green here, and yapf check failed on master.

Argh, sorry, I changed my setup before the last push, and forgot to reinstall the pre-commit hook. I'll submit a fix.

maximedenes:
> Yapf check is not running on PRs

It seems so. Can you open an issue? I can have a look at it too.

rth:
> It seems so. Can you open an issue? I can have a look at it too.

Maybe it a one off issue. It does run in other PRs.

maximedenes:
> It does run in other PRs.

I think I understand what is happening: the tests is run on all branches, but not on PRs. For PRs that correspond to branches of this repo, it's ok, but for other PRs it won't run. I'll try to fix it.

rth:
Makes sense, we need something like https://github.com/icubam/icubam/blob/1f0325d164f273c8151606ec1a3beb57ef974654/.github/workflows/pythontest-linux.yml#L4-L11 as done in other CI jobs.

PR #262, opened 2020-04-14
Run flake8 in CI

rprimet:
Running flake8 in CI, configured to report pyflakes-related errors only (we already use yapf for styling so a second tool seems unwarranted)

**Issues to consider**

- For now `predicu` is excluded
- A few errors where the fix is non-obvious need attention (see CI run results) ; any help appreciated


.github/workflows/flake8.yml
```diff
@@ -0,0 +1,24 @@
+name: run flake8 on source code
+
+on: push
```
rth, line 3: Let's move this to the same it CI job as yapf maybe? Otherwise starting one CI job per python linting tool will become a bit noisy in CI report.

icubam/db/test_synchronizer.py
```diff
@@ -217,7 +216,7 @@ def test_export_icus(self):
     with open("resources/test/icu2.csv") as csv_f:
       self.csv.sync_icus_from_csv(csv_f, False)
 
-    test_dir = tempfile.mkdtemp()
+    tempfile.mkdtemp()
```
rth, line 219: So visibly this temp folder is not actually used?

icubam/map_builder.py
```diff
@@ -1,8 +1,7 @@
-from absl import logging
```
rth, line 1: This has potential side effect, maybe add `# noqa` for now?

icubam/messaging/sms_sender.py
```diff
@@ -75,7 +75,7 @@ def get(config, sms_carrier=None):
   that is if the sender object is 'XXSender' the carrier name should be 'XX'
   """
   sms_carrier = sms_carrier or config.sms.carrier
-  cls_name = "{}Sender".format(sms_carrier)
+  "{}Sender".format(sms_carrier)
```
rth, line 78: If we don't use it, maybe there is no reason to do this formatting. Or just `str(sms_carrier)` to check that `__repr__` works?

icubam/predicu/__init__.py
```diff
@@ -1 +0,0 @@
-from .data import load_all_data, load_data_file, load_pre_icubam_data
```
rth, line 1: For `__init__` we can rather put these in,
```py
__all__ = ['load_all_data', 'load_data_file', 'load_pre_icubam_data']
```
which will fix linting. I'm not really sure we want to expose these at the module level, but in any case it's not a linting issue.

icubam/www/server.py
```diff
@@ -1,19 +1,16 @@
 import os.path
-from absl import logging
```
rth, line 2: Let's keep this, with `# noqa` it has side effects.

icubam/www/updater.py
```diff
@@ -1,5 +1,3 @@
-from absl import logging  # noqa: F401
```
rth, line 1: Also let's keep this.

icubam/db/test_synchronizer.py
rprimet: good catch!

.github/workflows/flake8.yml
rprimet: OK, will do

icubam/map_builder.py
rprimet: I think @dulacarnold asserted that it did not have side effects?

rth: > I think @dulacarnold asserted that it did not have side effects?

It does change the logging handler, see https://github.com/abseil/abseil-py/blob/06edd9c20592cec39178b94240b5e86f32e19768/absl/logging/__init__.py#L1167

but confirmation needed indeed.

dulacarnold: Maybe best to leave it in then, otherwise we might get inconsistent logging.

rth:
It will be fine, as long as we put it into the existing Python linting job https://github.com/icubam/icubam/pull/262#discussion_r408037971

maximedenes:
> It will be fine, as long as we put it into the existing Python linting job

Right, that looks indeed like the best option.

icubam/messaging/sms_sender.py
rprimet: I'll let others judge but it seems that the line can be wholly removed indeed?

icubam/predicu/__init__.py
rprimet: I'll just revert that change then, seeing as the CI flake8 configuration skips predicu anyway

PR #264, opened 2020-04-14
Changed upload endpoint + added preprocessor class and RoR preprocessor.

dulacarnold:


icubam/db/synchronizer.py
```diff
@@ -179,3 +188,41 @@ def export_all_bedcounts(self) -> TextIO:
...
+    # Remap columns and delete or default certain ones:
+    bc = bedcounts_df
+    bc.replace(self.ROR_COLUMNS_MAP)
+    del bc['ror_code']
+    bc['n_covid_occ'] = bc['n_covid_tot'] - bc['n_covid_free']
+    del bc['n_covid_tot']
+    bc[[
+      'n_covid_deaths', 'n_covid_healed', 'n_covid_refused',
+      'n_covid_transfered'
+    ]] = None
```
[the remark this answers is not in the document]
maximedenes: Edit: I saw that this is tested below, but it would be nice to have a short doc somewhere as a comment (if it does not exist yet).

icubam/backoffice/handlers/upload.py
```diff
@@ -29,7 +29,8 @@ def post(self):
     sync = synchronizer.CSVSynchcronizer(self.db)
```
olivierteboul, line 29: it seems like this is not the same name as in synchronizer anymore.

icubam/db/synchronizer.py
```diff
@@ -165,6 +165,15 @@ def sync_users_from_csv(self, csv_contents: TextIO, force_update=False):
     self.sync_users(users_df, force_update)
     return users_df.shape[0]
 
+  def sync_bedcounts_from_csv(self, csv_contents: TextIO, force_update=False):
```
olivierteboul, line 168: i am wondering if it would not be more appropriate to name those functions: `sync_icus`, etc. since we are already in the csvSynchronizer. And then the penultimate line becomes: `super().sync_users(...)`

icubam/db/test_synchronizer.py
```diff
@@ -71,53 +71,67 @@ def test_sync_bed_counts(self):
     bed_counts = []
     start_time = datetime.now(tz.tzutc())
 
-    # Inject a first set of bedcounts
-    bed_counts = self.gen_bed_counts(
+    # Inject a first set of bed_countss
```
olivierteboul, line 74: double s

icubam/www/handlers/db.py
```diff
@@ -83,3 +88,36 @@ def get(self, collection):
       os.remove(tmp_path)
     else:
       self.write(data.to_html())
+
+  @tornado.web.authenticated
```
maximedenes, line 92: Same question about the permission model here. Can any authenticated user post data?!

icubam/www/handlers/db.py
```diff
@@ -83,3 +88,36 @@ def get(self, collection):
...
+      file_path = os.path.join(self.upload_path, f"{time_str}-{file_name}")
+      try:
+        with open(file_path, "wb") as f:
+          f.write(file["body"])
+        logging.info(f"Received {file_path} from {self.request.remote_ip}.")
+      except IOError as e:
+        logging.error(f"Failed to write file due to IOError: {e}")
+    else:
+      logging.debug("POST API called with incorrect collection: {collection}.")
+      self.redirect(home.HomeHandler.ROUTE)
```
maximedenes, line 123: This should return an HTTP error code. If it is an API handler, I believe you shouldn't even redirect. It will make it easier for the client (whether an UI or test code) to detect and address the error.

Similar remark for other error cases above.

icubam/www/handlers/upload_csv.py
```diff
@@ -1,48 +0,0 @@
...
-  ACCESS = [store.AccessTypes.UPLOAD, store.AccessTypes.ALL]
-
-  def initialize(self, upload_path, config, db_factory):
-    self.upload_path = upload_path
-    super().initialize(config, db_factory)
-
-  @tornado.web.authenticated
-  def post(self):
-    file = self.request.files["file"][0]
-    file_name = clean_path(file["filename"])
```
maximedenes, line 39: I like that you no longer take the filename from the user, it is often dangerous from a security standpoint (even if it was sanitized here, you never know how these functions work when people play with null characters and the likes).

icubam/www/server.py
```diff
@@ -40,17 +39,16 @@ def make_routes(self):
     kwargs = dict(config=self.config, db_factory=self.db_factory)
     self.add_handler(home.HomeHandler, **kwargs)
     self.add_handler(home.MapByAPIHandler, **kwargs)
-    self.add_handler(db.DBHandler, **kwargs)
-    self.add_handler(VersionHandler, **kwargs)
-    self.add_handler(consent.ConsentHandler, **kwargs)
     self.add_handler(
-      upload_csv.UploadHandler, **{
+      db.DBHandler, **{
```
maximedenes, line 43: I don't know what the `db` endpoint was for before, didn't it make sense to keep it separated from the upload endpoint?

icubam/backoffice/handlers/upload.py
maximedenes: Does it mean this endpoint is not tested?

dulacarnold: shhhh

icubam/db/synchronizer.py
```diff
@@ -179,3 +188,41 @@ def export_all_bedcounts(self) -> TextIO:
...
+    """Sync bedcount CSVs from IdF RoR uplink."""
+    bedcounts_df = pd.read_csv(csv_contents)
+    col_diff = set(self.ROR_COLUMNS_MAP.keys()) - set(bedcounts_df.columns)
+    if len(col_diff) > 0:
+      raise ValueError(f"Missing columns in input data: {col_diff}.")
+
+    # Remap columns and delete or default certain ones:
+    bc = bedcounts_df
+    bc.replace(self.ROR_COLUMNS_MAP)
+    del bc['ror_code']
```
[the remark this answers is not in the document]
dulacarnold: I guess no need to delete, it will simply be ignored.  Remap just to make it easier to identify, if the incoming map changes just update that.

icubam/www/server.py
dulacarnold: It's our general API endpoint, so I think it's good to keep these together.

icubam/www/handlers/upload_csv.py
dulacarnold: yeah, I'm also glad we got rid of that.  I'm still a bit worried about pandas reading the provided CSV as-is, esp. because you can't force pandas to consider it as a file and not a URL (or I'm not sure how), so someone could potentially feed in a huuuuge URL and overload pandas' CSV reading.  I was thinking of going with csvlib here for that reason, @tgy or @rth any thoughts?  The user /is/ authenticated so there is also a certain level of trust.

icubam/www/handlers/db.py
dulacarnold: So normally only people with STATS UPLOAD or ALL tokens can post. @olivierteboul can you confirm?  We should also make the POST endpoint only accessible to ALL and UPLOAD keys (but STATS keys are a restricted set of people so also trusting here).

icubam/www/handlers/db.py
dulacarnold: Agreed, I won't lie, I wasn't sure how to do that.

icubam/backoffice/handlers/upload.py
```diff
@@ -29,7 +29,8 @@ def post(self):
...
     sync_fns = {
       'user': sync.sync_users_from_csv,
-      'icu': sync.sync_icus_from_csv
+      'icu': sync.sync_icus_from_csv,
+      'bedcounts': sync.sync_bedcount_from_csv
```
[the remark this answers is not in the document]
dulacarnold: No, normally the endpoint checks that they have a token with UPLOAD ALL or API access.  (and it really should be only ALL and UPLOAD).

icubam/db/synchronizer.py
```diff
@@ -165,6 +165,15 @@ def sync_users_from_csv(self, csv_contents: TextIO, force_update=False):
     self.sync_users(users_df, force_update)
     return users_df.shape[0]
 
+  def sync_bedcounts_from_csv(self, csv_contents: TextIO, force_update=False):
+    """Check that columns correspond, insert into a DF and synchronize."""
+    bedcounts_df = pd.read_csv(csv_contents)
+    col_diff = set(BC_COLUMNS) - set(bedcounts_df.columns)
+    if len(col_diff) > 0:
```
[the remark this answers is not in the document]
dulacarnold: I personally find this more clear, I am always weary of implicit 'None'/False in python.

icubam/db/synchronizer.py
dulacarnold: Will keep this explicit for now, I don't have a strong opinion but IIRC we agreed offline this was ok.

icubam/www/handlers/db.py
olivierteboul: This is a APIKeyProtectedHandler which means that the authentication is based on the API key. The key needs to exist, to be valid (not expired and active) and the restrictions must be one of   
 `store.AccessTypes.STATS, store.AccessTypes.ALL, store.AccessTypes.UPLOAD`

icubam/www/handlers/db.py
olivierteboul: I agree with Maxime. it is a post request so it should return an error instead of redirecting.
`self.set_status(4xx)` ?

olivierteboul: Plus I would start with checking it. this would prevent you from indenting a big block:
```
if collection != 'bedcounts':
  logging.error('something')
  self.set_status(400)
  return

...
```

dulacarnold: I want to first check the endpoint is correct, the block is for routing if we have other POST endpoints.

dulacarnold: But I've put a global check at the beginning for 404s.

PR #266, opened 2020-04-14
Smaller tokens + reset consent button

olivierteboul:
Sorry there are two orthogonal things in this PR:
- change the token to only content the two encoded ids 
- improved the authentication via token
- add a reset consent button in the user's page if the user has decline consent only.


icubam/www/handlers/update.py
```diff
@@ -27,47 +27,19 @@ def get_consent_html(self, user):
...
-      return None, None
-
-    return user, input_data
-
   async def get(self):
     """Serves the page with a form to be filled by the user."""
     user_token = self.get_query_argument(self.QUERY_ARG)
-    user, token_data = self.authenticate_from_token(user_token)
-    if user is None:
+    user, icu = self.decode_token(user_token)
```
dulacarnold, line 33: I just want to double-check that we're not making any auth sacrifices by getting rid of `authenticate_from_token`?

olivierteboul: no, actually i am removing duplicated code and moving authenticated_from_token to base.py while renaming it decode_token. It now returns the user and the icu objects.

This is better IMO since we use the same function here and in home.py (the map) instead of somewhat replicating the code.

Issue #237, opened 2020-04-12
Reduce the size of the Docker image


marccarre:
## Problem

The current docker image is rather large & heavy, which may:

- make it more costly (time, bandwidth, hosting requirements, etc.) to run `icubam`,
- make it less convenient for users to upgrade,
- make it slower to run CI,
- etc.

```console
$ docker images
REPOSITORY  TAG     IMAGE ID      SIZE
icubam      latest  50b08979695e  1.68GB
```

We should probably be able to reduce the image size by one order of magnitude given the size of these potential base images:

```console
$ docker images
REPOSITORY                          TAG       IMAGE ID      SIZE
python                              3.7-slim  74ac77e9873a  179MB
python                              3-slim    56930ef6f6a2  193MB
gcr.io/distroless/python3-debian10  latest    abd94187f786  52.2MB
gcr.io/distroless/python3           latest    be51572a0d67  49.6MB
```

## Suggestion

1. Reduce 3rd party dependencies as much as possible? (See also #235.)
2. ~Separate dev and prod Python dependencies?~ (Already `requirements.txt` and `requirements-dev.txt`.)
3. Separate build-time dependencies from run-time dependencies, using e.g. Docker multi-stage builds?
4. Using a lighter base image?
5. Following Docker best practices:

    1. have each `RUN` statement clean after itself, e.g.:

        ```console
        RUN apt-get update && apt-get install -y \
                ${PACKAGE} && \
            rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
        ```
    2. have a `.dockerignore` file to ensure we don't accidentally package more than required,

rprimet:
Yes, sounds nice!

(I believe the rationale for using the ubuntu base image is that previously, deployment would happen on a Ubuntu VM)

One added benefit is that, all dependencies being fetched on PyPI, we could do away with all the conda-related stuff in the Dockerfile?

Something to be mindful of, though, is that there was an implicit (except in the docs) dependency on sqlite >= 3.24.0 (which is newer than e.g. the one packaged in ubuntu 18.04 and was pulled by conda). But maybe this isn't even required now since the project moved from straight sql queries to sqlalchemy? cc @olivierteboul @dulacarnold?

As an aside maybe it'd be worth running the unit tests in the [docker packaging workflow](https://github.com/icubam/icubam/blob/master/.github/workflows/docker-package.yml) as well?

dulacarnold:
[...]

PR #270, opened 2020-04-15
Reduce Docker image size

pgraverdy:
Solves #237  (check issue for discussion)


docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # ARG required due to build option in docker-compose
 ARG ENV_MODE
 
-FROM ubuntu:18.04
+FROM python:3.7-slim-buster as builder
```
rth, line 4: Didn't we say that we use 3.8 in production in the end?

docker/Dockerfile
```diff
@@ -10,64 +10,48 @@ ENV TERM linux
...
-RUN wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
-RUN sh Miniconda3-latest-Linux-x86_64.sh -b
-ENV PATH /root/miniconda3/bin:$PATH
-
-# expose ports for flask app - both server and backoffice
-EXPOSE 8888 8890
+RUN apt-get -y update \
+    && apt-get install -y git \
+       wget \
+    && rm -rf /var/lib/apt/lists/*
```
rth, line 17: can also run a `apt-get clean -y` here.

docker/Dockerfile
```diff
@@ -10,64 +10,48 @@ ENV TERM linux
...
-RUN echo ".  /root/miniconda3/etc/profile.d/conda.sh" >> ~/.bashrc
-RUN conda create -n icubam python=3.8
-RUN conda activate icubam && pip install -r requirements.txt && pip install -e .
+RUN pip install -r requirements.txt && pip install -e .
+
+# clear pip cache
+RUN rm -rf ~/.cache/pip
+
+# make sure final image is clean
+FROM builder as launcher
```
rth, line 49: What does this do? Just curious.

docker/Dockerfile
pgraverdy: Yes, commit was not yet pushed, done now.

docker/Dockerfile
pgraverdy: If we don't do multi-stage (here having the "builder" intermediate layer) then the `rm` removes the files for the containers that will be started, but in terms of docker layers (hence image size) this has no effect (the layers where the files have been added is still in the "history" of layers). Building an intermediate image allow to solve this.

docker/Dockerfile
pgraverdy: Indeed. added on my side (although in this case there's no visible change on the image size)

rth:
Hopefully removing conda will also speed up CI a bit. Anything preventing from removing the draft status here?

pgraverdy:
@rth, indeed, removing Conda does reduce quite a lot the build time.

Below are the results of the command
```
time docker/scripts/docker_build.sh 1.0
```
On a VM on the Inria network, with no image in the local registry, for the following configurations
- Ubuntu 18.04 with conda install and pip install inside (current master docker file)
- Python 3.7 image with conda install and pip install inside
- Python 3.7 image with pip install directly

|  | | | |
| --- | --- | --- | --- |
| config | Ubuntu-conda  | Python-conda | Python |
| real |  2m45.524s | 2m43.217s |  1m20.772s|
| user | 0m0.384s |  0m0.324s| 0m0.184s |
| sys |  0m0.552s|  0m0.468s | 0m0.304s |

rth:
There is a merge conflict, would you mind merging master or rebasing manually @pgraverdy ?

docker/docker-compose.yml
```diff
@@ -0,0 +1,143 @@
+version: '3.2'
```
rth, line 1: Is it normal to have this new docker-compose.yaml here? I though this PR was only about reducing the image sizes?

pgraverdy: No, I think I messed up in the confilct resolution when merging master. Should not be there.

pgraverdy: @rth 
Hmm, since I had pushed the merge button on the PR a second time, there's another merge that piled up on top, and I personally do not feel confident enough to rewrite the history there. ok with you if I delete the unwanted file and push the commit or do you prefer to so a rewrite (all the commits will be squashed in the end).

rth: No of course, you can just remove the file and push another commit. We'll squash when merging by default.

PR #273, opened 2020-04-15
adding try catch

olivierteboul:


icubam/www/handlers/base.py
```diff
@@ -27,7 +27,15 @@ def decode_token(self, user_token: str) -> Tuple[store.User, store.ICU]:
...
 
-    userid, icuid = input_data
+    try:
+      userid, icuid = input_data
+    except Exception as e:
+      logging.error(f'Token is not a 2-tuple as expected: {e}')
+      userid, icuid = None, None
+      if isinstance(input_data, dict):
+        userid = input_data.get('user_id', None)
+        icuid = input_data.get('icu_id', None)
```
rth, line 37: Any tests we could add for this?

dulacarnold: It's a change in token format...

PR #275, opened 2020-04-15
FIX predicu: loading of combined bedcounts with public data

tgy:


rth:
> no comment on the deletion of

Probably fixing flake8 errors.

PR #276, opened 2020-04-15
two handlers solution for bug fix.

olivierteboul:
This PR follows PR 275. The problem was some old links, with tokens encoding a dict with several keys, were substituted by new links with only a tuple of two ids (user and icu) for space saving.

Now, when an old link would get it, things would break. Part of the solution was to fallback to the previous scheme in case of an bad token.

This bug also revealed another weird one. When calling self.render, the template would inject the current_user that is the result of get_current_user. Which would fail when trying to decode the cookie since the token had changed.

To make it more clear here, we split the handler that has no authentication whatsoever and the handler that use some authentication to update the bed counts.


icubam/www/handlers/update.py
```diff
@@ -51,6 +48,17 @@ async def get(self):
...
+  """Register new bedcounts coming from the form."""
+
+  ROUTE = updater.Updater.POST_ROUTE
+
+  def initialize(self, config, db_factory, queue):
+    super().initialize(config, db_factory)
+    self.queue = queue
+    self.updater = updater.Updater(self.config, self.db)
+
   @tornado.web.authenticated
```
maximedenes, line 62: How is this going to work, since `get_current_user` always returns `None`?

maximedenes:
Can you give a brief description of the bug this is fixing? At least for future reference.

icubam/www/handlers/update.py
maximedenes: Sorry, I got tricked by the collapsed diff. I understand now, please ignore my question.

icubam/www/handlers/update.py
```diff
@@ -28,18 +28,15 @@ def get_consent_html(self, user):
...
 
   async def get(self):
     """Serves the page with a form to be filled by the user."""
     user_token = self.get_query_argument(self.QUERY_ARG)
     user, icu = self.decode_token(user_token)
     if user is None or icu is None:
-      logging.error(f"Token authentication failed")
+      logging.error("Token authentication failed")
       self.clear_cookie(self.COOKIE)
       return self.set_status(404)
```
maximedenes, line 41: This should be a 403, not 404.

maximedenes:
> When calling self.render, the template would inject the current_user that is the result of get_current_user

Isn't this the expected behavior? https://www.tornadoweb.org/en/stable/guide/security.html#user-authentication

icubam/www/handlers/update.py
olivierteboul: actually we want to keep it a 404, as it the route does not exists... but we can change later

olivierteboul:
> > When calling self.render, the template would inject the current_user that is the result of get_current_user
> 
> Isn't this the expected behavior? https://www.tornadoweb.org/en/stable/guide/security.html#user-authentication

It is, but since things got a bit mixed up... the base handler does define a get_current_user. But we don't want to call it on the GET method. So we do not put a tornado.web.authenticated decorator.
But still this function get called in the self.render. So that is why i have split into two handlers. When we really want an authentication and when we dont.

maximedenes:
@olivierteboul I see, thanks for the explanation. I guess this discussion could be useful for future reference, if we want to improve a bit the authentication system at some point.

PR #277, opened 2020-04-15
Re-enable codecov

rth:
Re-enable codecov to be a little less strict. In particular, no matter the coverage report it should not fail the CI and thus prevent merge.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,21 @@
+comment: false
+
+coverage:
+  status:
+    project:
+      default:
+        # Commits pushed to master should not make the overall
+        # project coverage decrease by more than 4%:
+        target: auto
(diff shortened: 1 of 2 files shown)


rth:
OK I disabled codecov comments, because I think they are too noisy. Also increased the patch threshold to 50% so that unless the PR has fewer than ~25% of coverage it will not fail.
Reviewers should check that the coverage number in the CI status are not too bad. E.g. here,
Screenshot_2020-04-15 Re-enable codecov by rth · Pull Request #277 · icubam icubam

Note than even if a PR has failing codecov status, you can still merge it and it will be fine on master (at least for the next commit).

dulacarnold:
Great, this is nice, glad you got it removed from the comments too :)

PR #278, opened 2020-04-15
Faster python test CI

rth:
At attempt as slightly faster python tests CI by not using conda (same as for docker), and caching dependency wheels.


rth:
So actually some tests fail because sqlite version is 3.22.0 and we need >= 3.24.0. The error messages are DB related by otherwise in no way suggestive that there is a version issue. We should add a check for it.

rth:
It should be good to go. The CI job for python tests now takes ~1m40s, as opposed to ~2m20s on master.

PR #279, opened 2020-04-15
Add check for sqlite3 version in setup.py

rth:
Add check for minimal supported sqlite3 version in setup.py.

It the version is too low, things mostly work except for a few DB queries that return wrong results which is pretty bad. cf. https://github.com/icubam/icubam/pull/278#issuecomment-614182505

### DIFF
--- a/.github/workflows/docker-build-run.yml
+++ b/.github/workflows/docker-build-run.yml
@@ -1,4 +1,4 @@
-name: Build Docker image and test container
+name: docker build & test
 
 on:
   # Trigger the workflow on push or pull request,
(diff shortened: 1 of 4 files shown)


rth:
Checked that this works as expected in https://github.com/icubam/icubam/pull/278 (cf [CI log](https://github.com/icubam/icubam/pull/278/checks?check_run_id=589850315))

dulacarnold:
Very good catch!  I thought the sqlite version was intrinsically tied to the python version but it wasn't clear 100%, this makes things very explicit.  To be sure this is the version that support INSERT OR UPDATE? (UPSERT)?

rth:
>  To be sure this is the version that support INSERT OR UPDATE? (UPSERT)?

Yes, it is Sqlite 3.24 https://www.sqlite.org/lang_UPSERT.html

PR #281, opened 2020-04-15
Unauthorized error page

rpalmaotero:
This PR is related to #274 

- Adds an `ErrorHandler` in www server.
- Adds an `error.html `template in www server --similar to backoffice's login template-- explaining the situation and how to address it.


dulacarnold:
This is great, we're not leaking any actual error/exception information correct?

rpalmaotero:
Nope, the only message the user receives is that he/she hasn't visited an update form, and that it's required to do so in order to view the map.

dulacarnold:
LGTM, I'll let you sync and merge.

PR #282, opened 2020-04-15
Add preprocessing option to all_bedcounts HTTP query

tgy:


icubam/predicu/data.py
```diff
@@ -106,16 +106,19 @@ def load_bedcounts(
 
 
 def load_icubam(cached_data=None, api_key=None, icubam_host=None, clean=True):
-  if api_key is None or icubam_host is None:
-    raise RuntimeError("Provide API key and host to download ICUBAM data")
+  if cached_data is not None and 'raw_icubam' in cached_data:
```
dulacarnold, line 109: What happens here if there is no cached data for the API endpoint?

icubam/www/handlers/db.py
```diff
@@ -46,6 +48,9 @@ def initialize(self, config, db_factory):
   def get(self, collection):
     file_format = self.get_query_argument('format', default=None)
     max_ts = self.get_query_argument('max_ts', default=None)
+    should_preprocess = (
```
dulacarnold, line 51: Maybe sprinkle a couple comments as this might not be obvious to people down the road.

icubam/predicu/data.py
tgy: this `cached_data` dict is just a trick not to load the data from the API (`all_bedcounts` request) and use the data in the `cached_data` dict instead. see the call to `load_bedcounts` in the handler: i've added a comment about that there.

icubam/www/handlers/db.py
tgy: done.

PR #283, opened 2020-04-15
Docker compose scripts to help with new deployments.

sfehlandt:
# Docker tools for development
Added docker-compose scripts to build initial databases and to run the application mounting the code, so it is not necessary to rebuild the images:

## Changes
- Pass DB_SALT value to `app-backoffice` container in orginal docker-compose files
- Added `docker-compose-init-db.yml`and `build_fake_db.sh` to build the databases with docker
- Added `docker-compose-dev.yml`that mounts `scripts` and `icubam` folders on top of the fodlers that are copied in the Dockerfile
- Edited `docs/docker.md` accordingly

## How to init databases
```
docker-compose docker/docker-compose-init-db.yml build
docker-compose docker/docker-compose-init-db.yml up
docker-compose docker/docker-compose-init-db.yml down
```

## Develop with docker-compose
```
docker-compose docker/docker-compose-dev.yml build
docker-compose docker/docker-compose-dev.yml up -d
```
For example, if we make a change in the source code of the server (e.g. a template), we can see it simply restarting `app-server:`
```
docker-compose docker/docker-compose-dev.yml restart app-server
```


doc/docker.md
```diff
@@ -10,13 +10,36 @@ while the  `docker/docker-compose-proxy.yml` script launches the Nginx and CertB
...
+of the project (note: while both must be present to be mounted, only one will be used).  You can use the `docker/docker-compose-init-db.yml` script, see instructions below.
 - the configuration file (default name: ``config.toml``)
 
+### Use docker-compose for development
+
+#### Initalize DBs with docker-compose:
+```
+docker-compose docker/docker-compose-init-db.yml build
+docker-compose docker/docker-compose-init-db.yml up
+docker-compose docker/docker-compose-init-db.yml down
```
pgraverdy, line 28:
```suggestion
docker-compose -f docker/docker-compose-init-db.yml build
docker-compose -f docker/docker-compose-init-db.yml up
docker-compose -f docker/docker-compose-init-db.yml down
```

doc/docker.md
```diff
@@ -10,13 +10,36 @@ while the  `docker/docker-compose-proxy.yml` script launches the Nginx and CertB
...
+```
+
+#### Develop with docker-compose
+```
+docker-compose docker/docker-compose-dev.yml build
+docker-compose docker/docker-compose-dev.yml up -d
```
pgraverdy, line 34:
```suggestion
docker-compose -f docker/docker-compose-dev.yml build
docker-compose -f docker/docker-compose-dev.yml up -d
```

pgraverdy:
@sfehlandt 

Hi, 

Something wrt build path for docker and docker-compose that I think should be clearly explained in the doc.

I tried to run the command 
```
docker-compose -f docker/docker-compose-init-db.yml build
```
and it worked, but it was a side-effect of having setup in my environment ICUBAM_COMPOSE_CONTEXT to the project's root folder locally.  If setting this variable to empty, thus using the default value in the compose file, then I have the following result.
```
docker-compose -f docker/docker-compose-init-db.yml build
Building db-builder
ERROR: Cannot locate specified Dockerfile: ./docker/Dockerfile
```
If the `context` value is set to `"."` in the compose yml file, then docker-compose executes in the directory where the yml file is located (hence, "./docker/Dockerfile" is not available while "./Dockerfile" is).  I'm always launching docker-compose like
```
docker-compose -f docker/docker-compose-core.yml --project-directory . up
```
If we want to run the command without this `--project-directory` flag, then the default value in the compose file should be `".."`. However, this has an effect in production where we want the mounts' sources to be outside of the project's folder. This was the reason for using both project-directory and context (and keeping the same yml file).

Does it make sense ? Do you have another solution ?

pgraverdy:
@sfehlandt

Another discusion relating to dev vs. prod. 

Currently the Dockerfile on master uses Ubuntu:18.04 and Conda. However to reduce the size of the image and spped up build/testing, we are moving to the Python's buster-slim base image and will do away with Conda. 

Do you think the Docker image for development should extend the trimmed image and add by default a set of tools to properly dev/debug, or should it use another Dockerfile altogether ?

sfehlandt:
Hi @pgraverdy,

Yes, you are right. My compose assumes we have the `ICUBAM_COMPOSE_CONTEXT` variable set to `..`. I forgot to mention that. I added a suggested `.env` that works with all docker-compose scripts in the docker.md, and in that example I use `ICUBAM_COMPOSE_CONTEXT=..`

I tried to keep it versatile, maintaining your decision of keeping the `ICUBAM_COMPOSE_CONTEXT` configurable. If I understood correctly, the issue you mention with production could be solved keeping the `ICUBAM_COMPOSE_CONTEXT` as `..` but going up another level in mount points for the other resources.

About the docker.md, maybe I can add the `--project-directory .` to my commands in the instructions, with a note saying that you may omit that if you set `ICUBAM_COMPOSE_CONTEXT=..`

sfehlandt:
I think it is a good idea to split them. Maybe you can even use the alpine images for ultra lightweight production images and the buster-slim with other dev and debug tools (in the `requirements-dev.txt` right?).

What I have done in other projects is to have a `Dockerfile` optimized for production (as light as I can). And a Dockerfile-dev for development. I usually do not copy the whole code in that, as it is intended for development.
Basically I copy the requirements files and then run 'pip install -r requirements.txt'. Then in a docker-compose specific for development, I mount the whole source codo in the container. This way I can see changes with no need to rebuild the images. I also usually use Django or other frameworks that support "live changes refresh" when you run them in development mode. So I can see them as soon as I save the file...

pgraverdy:
Maybe regarding the build path, we can
- put the ICUBAM_COMPOSE_CONTEXT .env default values to .. 
- put in the docker.md the  --project-directory . 
- and put at the bottom the error case I described so that people that forget about the first two points may find the source of the error

Regarding the Dockerfile indeed better to have a dedicated Dockerfile as you propose (I'm from the node/angular world and indeed "live changes refresh" is quite convenient. BTW, you'll notice that the Dockerfile has changed (o more conda) due to the merge of the PR.

sfehlandt:
Great, I did as you said. About the first point I wasn't sure if you were talking about changing the default value in all docker-compose files. I did that (changed all the docker-compose files). But if you meant only the ones I added let me know and I can fix that.

docker/docker-compose-init-db.yml
```diff
@@ -0,0 +1,38 @@
...
+    volumes:
+      - type: bind
+        source: "${ICUBAM_CONFIG_PATH:-./resources/config.toml}"
+        target: /home/icubam/resources/config.toml
+      - type: bind
+        source: "${ICUBAM_COMPOSE_CONTEXT:-.}"
+        target: /home/icubam/db
+      - type: bind
+        source: "./build_fake_dbs.sh"
+        target: /home/icubam/build_fake_dbs.sh
```
pgraverdy, line 31: I had issues related to this related to the fact that the `project-directory` flag and the `context` are two different things that impact each other. With the provided setup and given the mount of the root of the project, I ended up with a folder named build_fake_dbs.sh at the root pf the project, and no db created.

When using docker-compose in `/home/pgraverdy/icubam` with project-directory value set to `.` and  build (`ICUBAM_COMPOSE_CONTEXT`) set to `..` then the build itself operates at  `/home/pgraverdy/icubam` (and not at `/home/pgraverdy/icubam/docker` if it was set to `.`) so the `./build_fake_dbs.sh` file is not found thus an empty folder is created by docker-compose.

Setting up ICUBAM_COMPOSE_CONTEXT to . allow the build part to process corretly, and since the script exists in the docker folder which is copied to the , we can just call it directly (no need to mount)


```suggestion
```

docker/docker-compose-init-db.yml
```diff
@@ -0,0 +1,38 @@
+version: '3.2'
+services:
+
+  # ICUBAM service
+  db-builder:
+    build:
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 7: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```

docker/docker-compose-dev.yml
```diff
@@ -0,0 +1,124 @@
...
+      - type: bind
+        source: ../scripts
+        target: /home/icubam/scripts
+    networks:
+      - icubam-network
+
+  # ICUBAM back-office service
+  app-backoffice:
+    build:
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 83: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
```

docker/docker-compose-dev.yml
```diff
@@ -0,0 +1,124 @@
...
+      - type: bind
+        source: ../scripts
+        target: /home/icubam/scripts
+    networks:
+      - icubam-network
+
+  # ICUBAM sms service
+  app-sms:
+    build:
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 46: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
```

docker/docker-compose-dev.yml
```diff
@@ -0,0 +1,124 @@
+version: '3.2'
+services:
+
+  # ICUBAM service
+  app-server:
+    build:
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 7: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
```

docker/docker-compose-core.yml
```diff
@@ -68,7 +68,7 @@ services:
   # ICUBAM back-office service
   app-backoffice:
     build:
-      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 71: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
```

docker/docker-compose-core.yml
```diff
@@ -4,7 +4,7 @@ services:
   # ICUBAM service
   app-server:
     build:
-      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 7: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
```

docker/docker-compose-core.yml
```diff
@@ -37,7 +37,7 @@ services:
   # ICUBAM sms service
   app-sms:
     build:
-      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
+      context: "${ICUBAM_COMPOSE_CONTEXT:-..}"
```
pgraverdy, line 40: See comment below about the mount
```suggestion
      context: "${ICUBAM_COMPOSE_CONTEXT:-.}"
```

docker/docker-compose-init-db.yml
sfehlandt: Ok, I accepted the changes

PR #291, opened 2020-04-16
Lookup FINESS data when importing ICUs in France

maximedenes:
If the country is FR, we lookup coordinates of the ICU in the FINESS
file.

### DIFF
--- a/icubam/db/synchronizer.py
+++ b/icubam/db/synchronizer.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from typing import TextIO
 
 import pandas as pd
+import json
 from absl import logging
 from dateutil import tz
 
@@ -13,6 +14,9 @@ from icubam.db import store
 # value, however columns that are used to create joins such as icu_name or
 # ICU_COLUMNS['name'] will throw an error if they are None or not aligned with
 # existing elements in the store.
+# The types here are not the real ones (see store.py for the schema), they are
+# just here to prevent pandas from messing with values, and delegate the actual
+# parsing to the SQL layer.
 ICU_DTYPE = {
   'name': 'string',
   'legal_id': 'string',
@@ -60,7 +64,7 @@ class StoreSynchronizer:
     else:
       logging.info("Admin Found!")
 
-  def sync_icus(self, icus_df, force_update=False):
+  def sync_icus(self, icus_df, finess_df, force_update=False):
     self.prepare()
 
     # pandas sometimes maps missing values (for example in CSV files) to NA
@@ -70,10 +74,22 @@ class StoreSynchronizer:
     for _, icu in icus_df.iterrows():
       icu_dict = icu.to_dict()
       icu_name = icu_dict["name"]
+      country = icu_dict["country"]
+      legal_id = icu_dict["legal_id"]
       region = icu_dict.pop('region', None)
       db_icu = self.db.get_icu_by_name(icu_name)
       if db_icu is not None and not force_update:
         continue
+
+      if legal_id is not None and finess_df is not None and (country == "FR"):
+        # We complete missing data using FINESS
+        print(f"Looking up {legal_id}")
+        row_df = finess_df.loc[finess_df["properties.nofinesset"] == legal_id]
+        coords = row_df["geometry.coordinates"].values[0]
+        if icu_dict["lat"] is None:
+          icu_dict["lat"] = coords[1]
+        if icu_dict["long"] is None:
+          icu_dict["long"] = coords[0]
 
       # Maybe create region first.
       if region is not None:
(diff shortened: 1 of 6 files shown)


icubam/db/synchronizer.py
```diff
@@ -4,12 +4,24 @@
...
 
 import pandas as pd
+import json
 from absl import logging
 from dateutil import tz
 
 from icubam.db import store
 
-ICU_COLUMNS = ['name', 'region', 'dept', 'city', 'lat', 'long', 'telephone']
+ICU_DTYPE = {
```
rth, line 13: Maybe just keep `ICU_COLUMNS` and make it a dict. Iterating over it will return keys as before.

icubam/db/synchronizer.py
```diff
@@ -147,13 +175,23 @@ def sync_bed_counts(self, bedcounts_df, user=None):
...
+  def __init__(self, store_db, finess):
+    super().__init__(store_db)
+    self.finess = finess
+
   def sync_icus_from_csv(self, csv_contents: TextIO, force_update=False):
     """Check that columns correspond, insert into a DF and sychronize."""
-    icus_df = pd.read_csv(csv_contents)
+    try:
+      finess_contents = open(self.finess, "r", encoding="utf-8")
+      finess_df = pd.json_normalize(json.load(finess_contents))
```
rth, line 186: Maybe add an assertion here, that finess_df.shape[1] > 1 or something similar?

icubam/db/test_synchronizer.py
```diff
@@ -142,7 +142,9 @@ def setUp(self):
       sqla.create_engine("sqlite:///:memory:", echo=True)
     )
     self.db = store_factory.create()
-    self.csv = synchronizer.CSVSynchcronizer(self.db)
+    self.csv = synchronizer.CSVSynchcronizer(
+      self.db, 'resources/finess.geojson'
```
rth, line 146: should we make a `resources/fr/finess.geojson` ?

The fact that this path works in tests, is it reliant on the folder from with pytest is started. E.g. does,
```
cd icubam/db/
pytest test_synchronizer.py
```
still works?

icubam/db/synchronizer.py
```diff
@@ -147,13 +175,23 @@ def sync_bed_counts(self, bedcounts_df, user=None):
 
 class CSVSynchcronizer(StoreSynchronizer):
   """Ingests CSV TextIO objects into datastore."""
+  def __init__(self, store_db, finess):
```
rth, line 178: There was the question whether we want this to generalize to other countries, in which case `finess=None` by default would be better.

icubam/db/test_synchronizer.py
dulacarnold: The problem is the json is huge, so we need this to fail gracefully for now.

icubam/db/synchronizer.py
dulacarnold: Agreed, it should be general, we should actually not use finess but icu_id or whatever the db uses as a variable name...

icubam/db/synchronizer.py
maximedenes: I tried initially, but we do other things than iterating over it IIRC, so I find it safer this way.

olivierteboul: all this information is available from store.py.  This is supposed to be the source of truth...
I would rather use as much store.py as possible to avoid duplicating the information and also avoiding things drifting appart.

icubam/db/synchronizer.py
olivierteboul: agreed

olivierteboul:
General comment:
I do not understand why we need to specify in the synchronizer the fields / columns of the csv and check that all the fields of the csv are exactly the expected ones? This seems to force a rigid behavior that forces those circumvolutions about finess numbers.

I would rather remove the if len(columns)>0: raise error and enable to update an ICU with whatever csv file you want, provided that:
- the columns are a subset of store.ICU.get_column_names()
- there is a name.
I can help with the PR if that is not clear.

maximedenes:
> I do not understand why we need to specify in the synchronizer the fields / columns of the csv and check that all the fields of the csv are exactly the expected ones? This seems to force a rigid behavior that forces those circumvolutions about finess numbers.

Is this changed by this PR, though?

icubam/db/synchronizer.py
maximedenes: If you look closer, you'll see that the types are not the same. The point of this type dictionary is to prevent pandas from doing random conversions when reading the CSV. In particular, if a value is missing in a row, pandas inserts a value which depends on the type it has inferred for the column (`NA` or `nan`, for example). It also messes with GPS coordinates, trying to round them.

So to summarize, the real schema is indeed defined by `store.py`. What we want here is to prevent pandas from messing with the data it passes to the store. I'll add a comment.

icubam/db/test_synchronizer.py
maximedenes: I made tests not rely on this file.

icubam/db/synchronizer.py
maximedenes: Done.

maximedenes:
A quick note: this just a first version, that we need in order to import data from the PACA region. The next step will be to generalize FINESS to arbitrary registries of medical services (from any country), and to import these in the DB rather than relying on external JSON data.

icubam/db/synchronizer.py
maximedenes: Done.

olivierteboul:
> A quick note: this just a first version, that we need in order to import data from the PACA region. The next step will be to generalize FINESS to arbitrary registries of medical services (from any country), and to import these in the DB rather than relying on external JSON data.

I understand, then i believe that if it is something ponctual like this, i would rather make a script that is specific instead of changing a code that is more generic, WDYT?

maximedenes:
> I understand, then i believe that if it is something ponctual like this, i would rather make a script that is specific instead of changing a code that is more generic, WDYT?

Yes, it is also a possibility. On the long term, I do think we'll want to change the generic code to consolidate the data, but I can probably keep it external for now.

On the other hand, this looks very similar to the ROR preprocessor that we added in the importer, doesn't it?

dulacarnold:
Hi @maximedenes , sorry I didn't give this a closer look before.  Indeed I agree with Olivier, let's keep the finess-specific lookups in a seperate script that will populate a CSV's legal_id field with the correct finess number (and perhaps the other values: city, lat, log etc.), and then we can use the standard import mechanism to just import that CSV instead of adding custom logic.  This is what we've done in the past for instance-specific imports. WDYT?

maximedenes:
> let's keep the finess-specific lookups in a seperate script

Ok, I agree it looks like a better solution until we add proper support for in-DB, country agnostic legal information.

> that will populate a CSV's legal_id field with the correct finess number

Note that this does is the other way around: it uses the provided finess number to lookup other details.

I'm closing this PR, will reopen one once I have a separate script.

PR #295, opened 2020-04-16
Make API access controls more fine-grained. Closes #293

dulacarnold:
Closes #293

PR #302, opened 2020-04-16
Fix small bug in csv import.

dulacarnold:


icubam/db/synchronizer.py
```diff
@@ -81,8 +82,12 @@ def sync_icus(self, icus_df, force_update=False):
 
       # If an ICU exists with the same name, update:
       if db_icu is not None:
-        manager = self._managers.get(db_icu.icu_id, self._default_admin)[0]
+        try:
+          manager = self._managers.get(db_icu.icu_id, self._default_admin)[0]
+        except TypeError:
+          manager = self._managers.get(db_icu.icu_id, self._default_admin)
```
rth, line 88: Si is the issue that self._managers.get returns None, and then  `None[0]` is a type error?

Maybe more explicitly then,
```py
manager = self._managers.get(db_icu.icu_id, self._default_admin)
if manager is not None:
   manager = manager[0]
```

icubam/www/handlers/db.py
```diff
@@ -50,6 +52,13 @@ def initialize(self, upload_path, config, db_factory):
 
   @tornado.web.authenticated
   def get(self, collection):
+    if self.current_user.acces_type not in self.GET_ACCESS:
+      logging.debug(
+        f"API called with incorrect access_type: {self.current_user.access_type}."
```
rth, line 57: That's from  https://github.com/icubam/icubam/pull/295, right, and so we need to first merge that one?

PR #305, opened 2020-04-16
fix predicu export script

tgy:


rth:
Should we add a unit test for the export function? E.g. monekypatch bedcount data with fake data as done in the plot tests? Currently it's not tested at all.

dulacarnold:
Codecov is blocking this, is that WAI?

rth:
It's blocking because `export_data` has no tests. We can still merge regardless but it would be nice to add a test.

tgy:
@rth I've added a test but the monkey patching does not seem to work.

could you check it out?

PR #308, opened 2020-04-16
Using a datalist to propose departments and cities in ICUs

olivierteboul:


icubam/backoffice/handlers/icus.py
```diff
@@ -49,12 +50,15 @@ def do_render(self, icu, error=False):
       regions = [e.region for e in self.current_user.managed_icus]
     regions.sort(key=lambda r: r.name)
 
+    icus = self.db.get_managed_icus(self.current_user.user_id)
+    depts = sorted(set([i.dept for i in icus]))
```
maximedenes, line 54: Can't we get the sorting done by the SQL layer, using an index?

olivierteboul: Understood. we probably could. However it would probably add quite some complexity on the store side to add a very adhoc feature that does not cost so much to be done in post processing (extract a field and sort)

icubam/backoffice/templates/icu.html
```diff
@@ -120,8 +130,23 @@ <h3 class="card-title">
...
+    $("input[data-bootstrap-switch]").each(function () {
+      $(this).bootstrapSwitch('state', $(this).prop('checked'));
+    })
+
+    function filInDataList (values, dataListId) {
+      var datalist = document.getElementById(dataListId);
+      values.forEach(function(item) {
+      var option = document.createElement('option');
+        option.value = item
+        datalist.appendChild(option)
```
rth, line 143: The current version is fine, but just to confirm you could fill those options with Jinja2 templating as well right?

icubam/backoffice/templates/icu.html
```diff
@@ -70,14 +70,24 @@ <h3 class="card-title">
...
+                  placeholder="{{ _('City') }}"
                   value="{% if icu.city %}{{icu.city}}{% end %}" required autofocus>
+                </input>
+                <datalist id="city_options"></datalist>
               </div>
 
-              <div class="col-25">
-                <label for="dept">{{ _('Department / State') }}</label>
-                <input type="text" id="dept" name="dept" class="form-control" placeholder="{{ _('Department') }}"
+              <div class="col-25 autocomplete">
```
rth, line 82: Does the `autocomplete` class do anything? Just curious.

icubam/backoffice/templates/icu.html
olivierteboul: absolutely. I was tempted to and to be honest i don t have a real reason to choose one rather than the other

icubam/backoffice/templates/icu.html
olivierteboul: no you are absolutely write. I started with another way and forgot to remove it. good catch. it is now removed.

PR #313, opened 2020-04-17
Added potential sentry SDK init.

dulacarnold:
* Add maybe_init_sentry call to initialise sentry if available.
* Add [tool.isort] to pyproject.toml to standardise python import sorting.


.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+force_single_line=True
```
rth, line 2: Let's rather do,
> You can also add your desired settings under a [tool.isort] section in your pyproject.toml file.

To avoid new config files. We should add it to `.pre-commit-config.yaml` -- I can make a PR.

icubam/backoffice/server.py
```diff
@@ -1,17 +1,28 @@
...
-  home, login, logout, users, tokens, icus, bedcounts, operational_dashboard,
-  messages, regions, maps, upload, consent
-)
+from absl import logging  # noqa: F401
+
 from icubam import base_server
+from icubam import utils
+from icubam.backoffice.handlers import bedcounts
+from icubam.backoffice.handlers import consent
+from icubam.backoffice.handlers import home
```
rth, line 15: What was wrong with the previous version? It was more concise.

Issue #312, opened 2020-04-17
Typos in department names


rth:
On the topic of typos in the department names. Here is what can happen when generating plots on production data,
```
Traceback (most recent call last):
  File "/home/rth/miniconda3/envs/icubam/lib/python3.8/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/rth/miniconda3/envs/icubam/lib/python3.8/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/rth/src/icubam/icubam/predicu/analytics_server.py", line 23, in generate_plots
    generate_plots(
  File "/home/rth/src/icubam/icubam/predicu/plot/__init__.py", line 159, in generate_plots
    plot(
  File "/home/rth/src/icubam/icubam/predicu/plot/__init__.py", line 121, in plot
    fig, tikzplotlib_kwargs = plot_fun(data=cached_data[data_source].copy())
  File "/home/rth/src/icubam/icubam/predicu/plot/barplot_deaths_per_dept_and_day.py", line 43, in plot
    facecolor=DEPARTMENT_COLOR[department],
KeyError: "Cotes-d'armor"
```
The key should have been `Côtes-d'Armor` there.

 - [ ] In the short term we should have a way to correct such typos as suggested by @tgy 
 - [ ] In the long term, as suggested by @olivierteboul  maybe everything should be purely in the DB and entered by users. https://github.com/icubam/icubam/pull/308 would help avoiding typos in that case. But then it's less clear how to provide this information to users who use only `icubam.predicu` with a CSV dump.

rth:
I think this is less of an issue with the new processing worflow, and since we added drop-down lists for departments.

Issue #311, opened 2020-04-16
Automate the pre-flight checklist for staging


dulacarnold:
Here is a list of pre-flight checks performed on staging before pushing to prod, and that we should automate as much as possible:

1. Check map loads.
2. Check update links load (taken from set of links in Bed Counts table in back office).
3. Check API endpoints load, and make sure STATS vs. UPLOAD keys have different endpoint auths.
4. Check API_KEY map access works with incognito window (no cookie set).
5. Make sure SMS server is alive (health link).
6. Make sure bedcounts/icus/users are populated.
....? Others?

Issue #301, opened 2020-04-16
Run some integration tests with production DB 


rth:
In some cases unit tests are not sufficient to ensure that components will work as expected.

For instance, generating analytics plots on fake data works, however when run on actual data it may fail e.g. https://github.com/icubam/icubam/issues/299  So in this particular case it would be good to run them on the actual data before putting in production.

This could be achieved as follows,

 - register an option to pytest such as `--icubam-config` (by default unused). It could could be used as follows,
   ```
   pytest icubam  --icubam-config=some_config.yaml
   ```
 - make a [pytest fixture](https://docs.pytest.org/en/latest/fixture.html) than when this option is provided, would copy the DB to some temporary location, and return a `Config` object to this locations.
 - some tests would then use this fixture, and when specified would run some integration tests on the copy of actual data. If it's not specified these tests are skipped.

IMO we shouldn't deploy periodic analytics plots generation in https://github.com/icubam/icubam/pull/222 without this, as the risk of failure is high.

Possibly related to production frontend tests @dulacarnold

rth:
Resolved in https://github.com/icubam/icubam/pull/314

Issue #263, opened 2020-04-14
Consider using pytest for tests in the future


rth:
Opening this issue for future reference to see how it could look like. There are definitely more important things to do now, and I know that @olivierteboul is not in favor.

So currently we use,
 - `inittest.TestCase`
 - in a few occasions `absltest.TestCase`
 - and for tornado tests `tornado.testing.AsyncHTTPTestCase`

if we were to migrate to pytest, the first would likely no longer be needed and the second one could be done with [pytest-tornado](https://github.com/eugeniy/pytest-tornado) plugin (see readme for more details).

Beside of being a bit less verbose, pytest allows e.g. parametric tests. For instance currently in https://github.com/icubam/icubam/blob/0ae313a5efe449830716bc8bfd331d82731f57f9/icubam/backoffice/server_test.py#L51
if one of the endpoints fail, the test exists without running the rest. With pytest parametrization, we could generate one test per endpoint that run independently. This is for instance already done in [`predicu/tests/test_plots.py`](https://github.com/icubam/icubam/blob/0ae313a5efe449830716bc8bfd331d82731f57f9/icubam/predicu/tests/test_plot.py#L21) for the same reason.

Again not pushing for this now, just FYI.

PR #314, opened 2020-04-17
Add mechanism for integration tests

rth:
Partially addresses https://github.com/icubam/icubam/issues/301

From the added documentation,

> Integrations tests can be run with,
>
> ```
> pytest -m integration --icubam-config=<config.toml path>
> ```
> This will detect the database in the dev section of the config file,
> make copy it to a temporary location and provide a new config object
> with the `integration_config` pytest fixture. It can be used as follows,
>
> ```py
> @pytest.mark.integration
> def test_some_integration(integration_config):
>     db_path = integration_config.db.sqlite_path
>  ```
>
> If ``--icubam-config`` key is not provided, the above test would be skipped.
> If ``-m integration`` option is not provided, both unit and integration tests
> are run (see [pytest documentation](https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests) for more details).

This would allow for instance to reproduce the https://github.com/icubam/icubam/issues/312 error (well after https://github.com/icubam/icubam/pull/222 is merged). Similarly it may help for https://github.com/icubam/icubam/issues/311 one could for instance get the correct AUTH token from the DB or the deployment hostname for production checks.

The actual DB passed via `--icubam-config` should never me modified, a backup is made, but just to be safe, maybe we should also open it in read-only mode.

Sorry this is all in pytest (https://github.com/icubam/icubam/issues/263), I don't know how to achieve this without it.


dulacarnold:
Ok great, if this is an argument to move everything to pytest, *AND* we are confident we won't break the tornado tests in the process, I am for moving everything to pytest, we however need to commit to migrate to a situation where everything works, we shouldn't move half the tests to pytest and need to disable the tornado tests in the meantime or somesuch.

Should we also reconsider tox in this case? There wase some argument to try out tox and if we're going to migrate maybe that's worth the extra step?  Not at all arguing for this but just bringing it back on the table in case.

doc/contributing.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Integration tests
+
+Integrations tests can be run with,
+
+```
+pytest -m integration --icubam-config=<config.toml path>
+```
+This will detect the database in the dev section of the config file,
+make copy it to a temporary location and provide a new config object
```
dulacarnold, line 26: "make a copy of it" or "copy it"?  or it's using Make to copy?

dulacarnold:
Also, does this mean we should have a more realistically populated DB to test out things such as plotting?

rth:
> Ok great, if this is an argument to move everything to pytest, AND we are confident we won't break the tornado tests in the process,

No it's not necessarily an argument for moving everything to pytest. The current situation is that we have classical unittests and unittests for tornado handlers, + some predicu tests that use pytest fixture. The whole things needs pytest to run  tests, but apart from that it doesn't require code migration one way or another. For tests in new modules, I personally would use plain test function (if there is not too much opposition from reviewers), but for existing ones they work as is and there is no need to change them. For simple changes, like https://github.com/icubam/icubam/pull/267 yes, we can migrate (or if someone wants to put in the effort), but I wouldn't migrate tornado handler tests, particularly as long as we have more pressing issues.

> Should we also reconsider tox in this case? There wase some argument to try out tox and if we're going to migrate maybe that's worth the extra step? 

I think this is orthogonal to pytest. tox basically reproduces part of CI locally, it's purely a dev tool. Personally I'm haven't seen the need to use it so for in this project, but if others developers find it useful we can certainly add it.

> Also, does this mean we should have a more realistically populated DB to test out things such as plotting?

Yes, this is intended to either test things locally on a real DB locally or before deployment, or we could try to make a real but heavily sanitized DB, say put it as a secret file in Github actions and run tests these tests on it in a separate job, only on upstream branches (not PRs). The questions is how to sync that integration DB with production. So maybe manual approach is still better.

PR #316, opened 2020-04-17
adding a covid switch button from covid+ to covid-

olivierteboul:
Probably not the best solution, but i am building two maps, and switching from one to the other.

It could be improved in a next PR, at least it works !


icubam/www/static/map.js
```diff
@@ -208,7 +214,17 @@ function plotMap(data, center, displayAllText, displayAllAltText) {
...
-  var centerControl = new CenterControl(centerControlDiv, map, displayAllText, displayAllAltText);
+  var centerControl = new CenterControl(
+    centerControlDiv, map, displayAllText, displayAllAltText, toggleAll);
   centerControlDiv.index = 1;
   map.controls[google.maps.ControlPosition.TOP_CENTER].push(centerControlDiv);
+
+  // Create the DIV to hold the control and call the CenterControl()
+  // constructor passing in this DIV.
+  var covidControlDiv = document.createElement('div');
+  const covidText = covid ? 'COVID+' : 'COVID-'
```
dulacarnold, line 225: Can we make it more clear that we are currently *seeing* the covid+/- map, maybe with "Currently: COVID+" or "Showing: COVID+"?

PR #323, opened 2020-04-18
Fix type errors and add mypy check

### LINKED ISSUE #180: Handle get_visible_bed_counts_for_user(user=None, force=False)
The case of `user=None` in `get_visible_bed_counts_for_user` when (`force=False`) should be more explicitly handled (by raising an exception). Currently it will produce an attribute error when accessing user fields.

See https://github.com/icubam/icubam/pull/167#discussion_r403485089

maximedenes:
I saw some code on master (probably now in prod) with typos in method names. I believe we really need something that catches such errors systematically before merge.

So this PR is fixing existing type errors on the code base and adding `mypy` to CI. The next step will be to add more type annotations, but the good thing is that it can be done incrementally.

Closes #180.

### DIFF
--- a/icubam/backoffice/handlers/base.py
+++ b/icubam/backoffice/handlers/base.py
@@ -2,7 +2,7 @@ import json
 import os.path
 import tornado.locale
 import tornado.web
-from typing import List, Dict, Union
+from typing import List, Dict, Union, Optional
 
 
 class BaseHandler(tornado.web.RequestHandler):
@@ -81,14 +81,16 @@ class BaseHandler(tornado.web.RequestHandler):
     """Given a store class, parse the body a dictionary."""
     result = dict()
     for col in cls.get_column_names():
-      value = self.get_body_arguments(col + '[]', None)
+      value: Union[
+        Optional[str],
+        List[str]] = self.get_body_arguments(col + '[]', strip=False)
       if not value:
         value = self.get_body_argument(col, None)
       if value is not None:
         result[col] = value
     return result
 
-  def format_list_item(self, item: Union[Dict, List]) -> list:
+  def format_list_item(self, item: Union[Dict, List]) -> Union[Dict, List]:
     """Prepare a dictionary representing a row of a table for display."""
     # TODO(olivier) improve this, too hard coded
     auto_links = {f'{k}_id': k for k in ['icu', 'user', 'region']}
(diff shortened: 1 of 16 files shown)


icubam/backoffice/handlers/base.py
```diff
@@ -81,14 +81,16 @@ def parse_from_body(self, cls) -> dict:
     """Given a store class, parse the body a dictionary."""
     result = dict()
     for col in cls.get_column_names():
-      value = self.get_body_arguments(col + '[]', None)
+      value: Union[
+        Optional[str],
+        List[str]] = self.get_body_arguments(col + '[]', strip=False)
```
rth, line 86: This one is strange. It have been inferred from `get_body_arguments` output. Is it well annotated?

icubam/db/store.py
```diff
@@ -756,8 +756,9 @@ def get_visible_bed_counts_for_user(
     kargs are passed to get_latest_bed_counts_for_icus() method for additional
     filtering, e.g. max_date.
     """
-    # TODO: handle user=None here (#180)
     user = self.get_user(user_id)
+    if not user:
+      raise ValueError(f"Cannot find user with id {user_id}")
```
rth, line 761: Please add "closes #180" to the issue description to auto-close.

maximedenes: Right, done, thanks!

icubam/backoffice/handlers/base.py
maximedenes: To be honest, I think this code would deserve some explanations, not just type annotations ;)

maximedenes:
@rth Thanks for the feedback, yes, I'll add it to the pre-commit hook. Why `--follow-imports skip`, though? The doc says "We do not recommend using skip unless you know what you are doing: while this option can be quite powerful, it can also cause many hard-to-debug errors.".

rth:
> Why --follow-imports skip, though?

You are right, copied that from a different projcet. It's not needed.

icubam/backoffice/handlers/base.py
maximedenes: I don't know how type inference in `mypy` works, but at least this cannot be guessed locally, since we then assign `get_body_argument` (note the singular). They return different types.

It's not trivial for me to figure out the intention behind this code, so maybe @olivierteboul can give some explanations, then I can add a comment and improve the annotations if possible.

rth: OK, I guess we can leave it as is for now and improve later.

maximedenes:
To anybody merging this PR, please DO NOT SQUASH it. If we ever revert the addition of mypy to CI and/or the pre-commit hook, we probably don't want to get the bugs back with it :)

Issue #296, opened 2020-04-16
Language dependent resources


rth:
To keep ICUBAM generic and not specific to France we should avoid any hard-coded ressources. In particular, we should move [icubam/predicu/data](https://github.com/icubam/icubam/tree/master/icubam/predicu/data) maybe to some language specific path.

One thing I don't understand about the current resources folder, is that tests use relative paths e.g. `resources/test.toml` meaning that if python is started from a different folder than ICUBAM root directory, it will not work.

For instance,
```
cd /tmp/
pytest --pyargs icubam -k config
```
currently fails with
```
E       Exception: Couldn't find INI file: resources/test.toml
```
That's OK as long as it's only test config files, but that would not be acceptable with actual data ressources.

I think we might want to have,
 - `resources/` with config files as we currently have.
 - `icubam/data/fr` or `icubam/resources/fr/` for language specific data files that we want to vendor inside the project. Then include those in `package_data` in `setup.py` 

WDYT? cc @olivierteboul

dulacarnold:
Yeah, I'm not sure how to set test resource file paths otherwise, any suggestions?

rth:
For test data, I think we should put it under `icubam/resources/test` or `icubam/data/test`, define e.g. `TEST_DATA_PATH` with respect to `__file__` in `icubam/data/__init__.py` and use paths with respect to this base directory in tests.

PR #329, opened 2020-04-18
Fix CSV import of bedcounts and add test

maximedenes:
The importer was calling `sync_users` instead of `sync_bedcounts`. We
also fix parsing of dates and add a test for this importer.

### DIFF
--- a/icubam/db/synchronizer.py
+++ b/icubam/db/synchronizer.py
@@ -154,6 +154,7 @@ class StoreSynchronizer:
         raise ValueError(
           "Timestamps must be in UTC, got {}".format(bc['create_date'].tzinfo)
         )
+
       item = bc.to_dict()
       # Replace icu_name with corresponding ID:
       item['icu_id'] = db_icus[item['icu_name']].icu_id
@@ -168,7 +169,9 @@ class StoreSynchronizer:
 
 class CSVSynchronizer(StoreSynchronizer):
   """Ingests CSV TextIO objects into datastore."""
-  def sync_icus_from_csv(self, csv_contents: TextIO, force_update=False):
+  def sync_icus_from_csv(
+    self, csv_contents: TextIO, force_update=False
+  ) -> int:
     """Check that columns correspond, insert into a DF and sychronize."""
     icus_df = pd.read_csv(csv_contents, dtype=ICU_DTYPE)
     col_diff = set(ICU_COLUMNS) - set(icus_df.columns)
@@ -177,7 +180,9 @@ class CSVSynchronizer(StoreSynchronizer):
     self.sync_icus(icus_df, force_update)
     return icus_df.shape[0]
 
-  def sync_users_from_csv(self, csv_contents: TextIO, force_update=False):
+  def sync_users_from_csv(
+    self, csv_contents: TextIO, force_update=False
+  ) -> int:
     """Check that columns correspond, insert into a DF and synchronize."""
     users_df = pd.read_csv(csv_contents)
     col_diff = set(USER_COLUMNS) - set(users_df.columns)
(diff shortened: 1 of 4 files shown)


icubam/db/synchronizer.py
```diff
@@ -192,7 +192,10 @@ def sync_bedcounts_from_csv(self, csv_contents: TextIO, force_update=False):
     col_diff = set(BC_COLUMNS) - set(bedcounts_df.columns)
     if len(col_diff) > 0:
       raise ValueError(f"Missing columns in input data: {col_diff}.")
-    self.sync_users(bedcounts_df, force_update)
+    bedcounts_df['create_date'] = pd.to_datetime(
+      bedcounts_df['create_date']
+    ).dt.tz_localize("Europe/Paris").dt.tz_convert(tz.tzutc())
```
dulacarnold, line 197: I had tried to set the TZ in the import script to not hard-code it her.  Could also pass in the tz string, but probably best to not fix it to Europe/Paris.

maximedenes: Good point. What I don't understand in the previous code is that the datetime does not seem to be parsed here. Doesn't it always fail on any result from `read_csv`?

So maybe I should keep the parsing but not the time conversion. Adjustment of TZ should be done by a preprocessor.

maximedenes: Should be better now. I struggled a bit because pandas and dateutils don't use the same object to represent UTC, it seems...

icubam/db/synchronizer.py
```diff
@@ -186,13 +190,19 @@ def sync_users_from_csv(self, csv_contents: TextIO, force_update=False):
...
-  def sync_bedcounts_from_csv(self, csv_contents: TextIO, force_update=False):
+  def sync_bedcounts_from_csv(
+    self, csv_contents: TextIO, force_update=False
+  ) -> pd.DataFrame:
     """Check that columns correspond, insert into a DF and synchronize."""
-    bedcounts_df = pd.read_csv(csv_contents)
+    bedcounts_df = pd.read_csv(
+      csv_contents,
+      parse_dates=BC_DATES_COLUMNS,
+      date_parser=lambda x: pd.to_datetime(x, utc=True)
```
dulacarnold, line 200: With something like
``` d["create_date"] = d.datetime.dt.tz_localize("Europe/Paris").dt.tz_convert(tz=tz.tzutc())``` 
you should be able load the time with the correct locale and then convert to UTC. This check will be compatible with the previous validation method in Synchronizer as well.

File: icubam/db/synchronizer.py [hunk 3]
```diff
@@ -31,6 +31,7 @@
   'n_covid_deaths', 'n_covid_healed', 'n_covid_refused', 'n_covid_transfered',
   'create_date'
 ]
+BC_DATES_COLUMNS = ['create_date']
```
dulacarnold, line 34: Maybe put this constant inside the CSV class?  Not sure about my suggestion...

resources/test/bedcounts.csv
```diff
@@ -0,0 +1,2 @@
+icu_name,n_covid_occ,n_covid_free,n_ncovid_occ,n_ncovid_free,n_covid_deaths,n_covid_healed,n_covid_refused,n_covid_transfered,create_date
+hopital3,11,12,13,14,15,16,17,18,2020-04-16T23
```
dulacarnold, line 2: What is 16T23?

maximedenes: It's 11pm on the 16th.

icubam/db/synchronizer.py
maximedenes: Sorry, what's the difference with what I had previously?

icubam/db/synchronizer.py
maximedenes: This has to stay in sync with columns (if we change names, or add new date columns), so I think it is safer here.

resources/test/bedcounts.csv
dulacarnold: Ah ok, is this a common format? Is that what we want in our tests?

maximedenes: No strong opinion, I changed it to something more common.

icubam/db/synchronizer.py
maximedenes: I went back to the previous solution and added a timezone parameter to avoid hardcoding it.

maximedenes:
@dulacarnold This should be ready. I have a script that calls this importer after preprocessing FINESS data, that I'll submit as a subsequent PR.

icubam/db/test_synchronizer.py
```diff
@@ -236,3 +236,11 @@ def test_export_icus(self):
       str_buf.splitlines(),
       len(icus) + 1, "Number of rows should be the same."
     )
+
+  def test_import_bedcounts(self):
+    with open("resources/test/icu2.csv") as csv_f:
+      self.csv.sync_icus_from_csv(csv_f, False)
+    with open("resources/test/bedcounts.csv") as csv_f:
```
rth, line 243: In the long term I we should change the folder where these files are stored https://github.com/icubam/icubam/issues/296#issuecomment-616095233

PR #330, opened 2020-04-18
New map UI and better covid +/- toggling.

dulacarnold:
image


icubam/www/static/map.js
```diff
@@ -217,14 +218,21 @@ function plotMap(data, center) {
   var centerControl = new CenterControl(
     centerControlDiv, map, displayAllText, displayAllAltText, toggleAll);
   centerControlDiv.index = 1;
-  map.controls[google.maps.ControlPosition.TOP_CENTER].push(centerControlDiv);
-
+  map.controls[google.maps.ControlPosition.BOTTOM_CENTER].push(centerControlDiv);
+  var titleDiv = document.getElementById('title-box-orig')
```
olivierteboul, line 222: maybe you could simple `$(#title_id).text(...)` or even only change the +/- directly in the html via jquery or in pure js with innerHTML instead of creating two titles

dulacarnold:
Committing as is, and we will optimise this a bit later.

PR #333, opened 2020-04-19
FIX Disable the check of None user_id in get_visible_bed_counts_for_user

rth:
Re-opens https://github.com/icubam/icubam/issues/180

The fix in https://github.com/icubam/icubam/issues/180 was not sufficient as it also needs to account for `force=True`. For now just revert it.

### DIFF
--- a/icubam/db/store.py
+++ b/icubam/db/store.py
@@ -755,8 +755,14 @@ class Store(object):
 
     kargs are passed to get_latest_bed_counts_for_icus() method for additional
     filtering, e.g. max_date.
+
+    Args:
+      user_id: run query for a given user
+      force: run query as admin irrespective of the user_id
     """
     user = self.get_user(user_id)
+    if not force and user is None:
+      raise ValueError("A non null user_id must be provided")
     if force or user.is_admin:
       return self._get_bed_counts_for_icus(None, latest=True, **kargs)
     else:
--- a/icubam/db/store.py
+++ b/icubam/db/store.py
@@ -757,8 +757,6 @@ class Store(object):
     filtering, e.g. max_date.
     """
     user = self.get_user(user_id)
-    if not user:
-      raise ValueError(f"Cannot find user with id {user_id}")
     if force or user.is_admin:
       return self._get_bed_counts_for_icus(None, latest=True, **kargs)
     else:
--- a/icubam/db/store.py
+++ b/icubam/db/store.py
@@ -764,7 +764,7 @@ class Store(object):
     if force:
       return self._get_bed_counts_for_icus(None, latest=True, **kargs)
     elif user is None:
-      raise ValueError("a non null user_id must be provided")
+      raise ValueError(f"Cannot find user with id {user_id}.")
     elif user.is_admin:
       return self._get_bed_counts_for_icus(None, latest=True, **kargs)
     else:
--- a/icubam/db/store.py
+++ b/icubam/db/store.py
@@ -761,9 +761,11 @@ class Store(object):
       force: run query as admin irrespective of the user_id
     """
     user = self.get_user(user_id)
-    if not force and user is None:
-      raise ValueError("A non null user_id must be provided")
-    if force or user.is_admin:
+    if force:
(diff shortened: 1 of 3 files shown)


dulacarnold:
We need integration tests, this only triggered because I checked by hand on the staging server...  Actually we need: TornadoTests on the API, that would have found it, and also some automated tests on staging...  But good fix, I was digging through trying to figure out when this broke :)

maximedenes:
At least the call to `user.is_admin` should be guarded.

rth:
Good thing it wasn't in production. I agree we need better tests.

> At least the call to user.is_admin should be guarded.

Yes, let me do something a bit better that would also keep mypy happy.

maximedenes:
What's the spec of `force`?

maximedenes:
Probably we should:
- guard the `user.is_admin` call
- make sure the user is real in the `else` branch
- add a brief description of the spec of `force` as a comment

dulacarnold:
> What's the spec of `force`?

@olivierteboul can comment better, but force was basically there it to allow us to continue operating while proper authentication was put in place.  Because the API handles authentication differently (no associated user), force is the only way currently to get the store to provide necessary data, and allow the API to manage auth through key-based mechanisms.

icubam/db/store.py
```diff
@@ -755,11 +755,17 @@ def get_visible_bed_counts_for_user(
...
     user = self.get_user(user_id)
-    if not user:
-      raise ValueError(f"Cannot find user with id {user_id}")
-    if force or user.is_admin:
+    if force:
+      return self._get_bed_counts_for_icus(None, latest=True, **kargs)
+    elif user is None:
+      raise ValueError("a non null user_id must be provided")
+    elif user.is_admin:
       return self._get_bed_counts_for_icus(None, latest=True, **kargs)
```
rth, line 769: OK I refactored a bit. Now it's a bit verbose (and there is one repetition) but mypy is able to correctly narrow down the types of `user` from `Optional[int]` to `int` after the None check.

Previously with,
```
if user is not None and not force:
    raise ValueError
```
the type after the check could still be None.

icubam/db/store.py
```diff
@@ -755,11 +755,17 @@ def get_visible_bed_counts_for_user(
...
+      force: run query as admin irrespective of the user_id
     """
     user = self.get_user(user_id)
-    if not user:
-      raise ValueError(f"Cannot find user with id {user_id}")
-    if force or user.is_admin:
+    if force:
+      return self._get_bed_counts_for_icus(None, latest=True, **kargs)
+    elif user is None:
+      raise ValueError("a non null user_id must be provided")
```
dulacarnold, line 767: Maybe use capitals and punctuation while we're at it.

maximedenes: Also, maybe `user_id` is not actually null be just the user can't be found, right?

icubam/db/store.py
```diff
@@ -755,11 +755,17 @@ def get_visible_bed_counts_for_user(
...
+      user_id: run query for a given user
+      force: run query as admin irrespective of the user_id
     """
     user = self.get_user(user_id)
-    if not user:
-      raise ValueError(f"Cannot find user with id {user_id}")
-    if force or user.is_admin:
+    if force:
+      return self._get_bed_counts_for_icus(None, latest=True, **kargs)
+    elif user is None:
```
dulacarnold, line 766: Could we just do `elif user is None and not force:`?

icubam/db/store.py
```diff
@@ -755,11 +755,17 @@ def get_visible_bed_counts_for_user(
...
+      raise ValueError("a non null user_id must be provided")
+    elif user.is_admin:
```
dulacarnold, line 768:
```suggestion
    elif user.is_admin or force:
```

icubam/db/store.py
dulacarnold:
```suggestion
      raise ValueError("A non null user_id must be provided.")
```

icubam/db/store.py
maximedenes: Don't we already know `force == false` at that point?

PR #335, opened 2020-04-19
Add script to preprocess ICUs CSV files based on FINESS data

maximedenes:
This PR replaces #291 with an external script which preprocesses ICU CSV files before import. Based on the provided FINESS data, it automatically adds GPS coordinates and city information. It is useful e.g. when adding all ICUs from a new region.


maximedenes:
Note: the data files are not provided because they are large and specific to France.

PR #336, opened 2020-04-19
Disable isort pre-commit hook

rth:
Following earlier discussion due to conflicts between yapf and isort.
Manual usage is still encouraged.


dulacarnold:
:( 👋 😢

PR #338, opened 2020-04-19
New decorator to allow specific HTTP status instead of redirection when not authenticated

olivierteboul:


icubam/www/handlers/base.py
```diff
@@ -75,6 +76,21 @@ def get_user_locale(self):
       return tornado.locale.get(locale_code)
 
 
+def authenticated(func=None, *, code=503):
```
rth, line 79: So this just returns an HTTP error code without rendering any template right. Which is what we want for API calls but not user facing endpoints.

Maybe call this `authenticated_api` or put a sentence in the docsting to indicate that it should not be used with non API endpoints?

icubam/www/handlers/db.py
```diff
@@ -50,7 +50,7 @@ def initialize(self, upload_path, config, db_factory):
       self.db.get_visible_bed_counts_for_user, user_id=None, force=True
     )
 
-  @tornado.web.authenticated
+  @base.authenticated(code=503)
```
rth, line 53: Could we add a tests and assert that code?

rth:
The inner wrapper is never run in tests, according to codecov. I think we should add a tests to run it on at least one of the endpoints where it's used.

PR #340, opened 2020-04-20
Fix f-string in conftest.py

rth:
Minor. Otherwise it would create a `{db_path_new}` file in project dir, instead of in a temporary location.


conftest.py
```diff
@@ -32,7 +32,7 @@ def integration_config(request, tmpdir_factory):
 
   db_path_new = str(tmpdir_factory.mktemp('tmpdir_factory').join("icubam.db"))
 
-  exit_code = call(['sqlite3', str(db_path), '.backup {db_path_new}'])
+  exit_code = call(['sqlite3', str(db_path), f'.backup {db_path_new}'])
```
maximedenes, line 35: Wouldn't it be safer to escape, e.g. `f".backup '{db_path_new}'"`, something like that? For example if the path contains spaces.

rth: Thanks. Quoting path could be useful indeed.

rth:
@maximedenes I addressed your comments.

Also added an assertion to check that the copied DB does exist, otherwise it would just create a new empty DB, which is what's happening until this PR is merged.

PR #342, opened 2020-04-20
Adding tests to authenticated decorator

olivierteboul:


rth:
Could you please master in? I can't do it in Github UI due to "This branch cannot be rebased safely".

PR #343, opened 2020-04-20
adding emails for every users (optional)

olivierteboul:


dulacarnold:
Is there an easy way to do e-mail validation with your frontend framework?  IF so might be worth adding...

PR #344, opened 2020-04-20
Wipe using SQLAlchemy API + reset admin option

rprimet:
This should have been two separate PRs but, oh well...

Rewrite the wipe script using SQLAlchemy to gain some future-proofing + add the option to reset admin username/password (may be useful on staging server)

### DIFF
--- a/scripts/wipe_sensitive_info_db.py
+++ b/scripts/wipe_sensitive_info_db.py
@@ -6,22 +6,29 @@ Warning: modifies data in-place!
 from absl import app
 from absl import flags
 import click
-import sqlite3
+import os
+from sqlalchemy import create_engine
 
+from icubam.db.store import StoreFactory
 from icubam.db.wipe import wipe_db
 
 flags.DEFINE_boolean("force", False, "Do not prompt user confirmation.")
 flags.DEFINE_boolean(
   "keep_beds", False, "Whether to keep bed occupation data."
 )
+flags.DEFINE_boolean("reset_admin", False, "Reset admin information")
+flags.DEFINE_string("admin_email", None, "Admin email (if resetting admins)")
+flags.DEFINE_string("admin_pass", None, "Admin password (if resetting admins)")
 flags.DEFINE_string("filename", None, 'DB file name.')
 flags.mark_flag_as_required('filename')
 FLAGS = flags.FLAGS
 
 
-def wipe_db_path(path, keep_beds):
-  conn = sqlite3.connect(path)
-  wipe_db(conn, keep_beds)
+def wipe_db_path(path, keep_beds, reset_admin, admin_email, admin_pass):
+  store_factory = StoreFactory(create_engine("sqlite:///" + path))
+  wipe_db(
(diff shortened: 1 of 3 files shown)


scripts/wipe_sensitive_info_db.py
```diff
@@ -30,7 +37,10 @@ def main(argv):
       "WARNING: THIS WILL WIPE THE DATABASE IN-PLACE. CONTINUE?", err=True
     ):
       return
-  wipe_db_path(FLAGS.filename, FLAGS.keep_beds)
+  wipe_db_path(
+    FLAGS.filename, FLAGS.keep_beds, FLAGS.reset_admin, FLAGS.admin_email,
+    os.environ.get("ICUBAM_ADMIN_PASS", FLAGS.admin_pass)
```
rth, line 42: This doesn't seem to be documented in the script help.

rprimet:
@rth regarding the in-place wipe I thought it was generally safe as it is meant to wipe a copy of a prod backup on the staging server, and it needs an explicit flag. But if you feel that's not enough I can implement either of your suggestions.

scripts/wipe_sensitive_info_db.py
rprimet: I'm not sure how the script help works -- isn't it autogenerated? Or did you mean that the env variable override is not documented?

rth:
> regarding the in-place wipe I thought it was generally safe as it is meant to wipe a copy of a prod backup on the staging server, and it needs an explicit flag.

Sounds good then.

rth:
Approving, but as mentioned above a second review would be nice.

rprimet:
Yep, will wait for a second pair of eyes :-)

scripts/wipe_sensitive_info_db.py
```diff
@@ -6,22 +6,32 @@
...
+from icubam.db.store import StoreFactory
 from icubam.db.wipe import wipe_db
 
 flags.DEFINE_boolean("force", False, "Do not prompt user confirmation.")
 flags.DEFINE_boolean(
   "keep_beds", False, "Whether to keep bed occupation data."
 )
+flags.DEFINE_boolean("reset_admin", False, "Reset admin information")
+flags.DEFINE_string("admin_email", None, "Admin email (if resetting admins)")
+flags.DEFINE_string(
```
maximedenes, line 21: In the typical use case, what's this admin password? Is it a fake password for the wiped DB, or something more real/sensitive? If the former, leaking the env variable into it feel strange, if the latter, we should probably prompt for it rather than take it as a command line argument.

rprimet: I had two use cases in mind: quick checks on a dev machine (fake password) and staging server (stronger password even though there should be no sensitive info in the DB, it would probably be stored in the CI secrets hence the environment variable)

maximedenes: Ok, fair enough. We can hope users know what they are doing :)

Issue #284, opened 2020-04-16
Improve visualization page in backoffice


rth:
Just opening an issue to track what should be done on visualizations to allow displaying up to data information to decision makers following discussion with @dulacarnold @tgy .

**Short term**
 Currently we have the ability to render any plots in some folder into the `/bo/operational-dashboard` page.
 We would need to,
 - [x] merge https://github.com/icubam/icubam/pull/222 to re-generate predicu plots periodically in the backoffice server  and check that it doesn't explode memory on the server due to unclosed objects in matplotlib
 - [x] at least add descriptions to predicu plots and decide which ones we should show
 - [x] check results consistency and apply some filtering on the raw data if needed (as done for off-line reports)
 - [x] add token based authentication for this page, that should only give access to this page and nothing else

**Mid term**

Mid term we might want to move the plot generation to separate webservice to make it a bit more interactive and not need to generate plots for a cartesian product of all possible options (regions, cities, time durations, etc). If we go this road, we can certainly get rid of the bokeh figure and close https://github.com/icubam/icubam/issues/154

rth:
A status update on this is that after https://github.com/icubam/icubam/pull/350 is merged, we can display all plots without any formatting at http://localhost:8890/bo/operational-dashboard the callback to generate plots will also work (put at least 4 hours periodicity to avoid memory issues on the server).

Now we focus only on latest @fplandes 's plots "flux_dept_lits_dept_visu_4panels" that are done by region and department. 

Next steps are,
 - [ ] add token based authentication (still)
 - [ ] improve layout in particular we want to be able to select a region or department and see the plot for it. This means we need a way to generate the correct path to the corresponding image (e.g. `/bo/static/extra-plots/flux-lits-dept-Ardennes.png`) for a given Region / departments. Issues there are non ascii characters in names, related to https://github.com/icubam/icubam/issues/312.
 - [ ] we may want to split each plot in two: for covid+ and covid- information and display them on several tabs.
 - [ ] double check information on the plots. Plot generation currently produces a lot of verbose information (e.g. about departments missing) which might need investigation.

rth:
Closing as resolved.

PR #350, opened 2020-04-20
Working on integration of predicu plots into the dashboard

rth:
Partially addresses https://github.com/icubam/icubam/issues/284

### DIFF
--- a/icubam/backoffice/templates/operational-dashboard.html
+++ b/icubam/backoffice/templates/operational-dashboard.html
@@ -85,7 +85,7 @@
   {% for fig_name in row %}
   {% if fig_name is not None %}
       <div class="col-sm-5 figure">
-         <img src="/bo/static/extra-plots/{{ fig_name }}.png" />
+        <img src="/{{ backoffice_root }}/static/extra-plots/{{ fig_name }}.png" />
   
       </div>
   {% end %}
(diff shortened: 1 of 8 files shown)


icubam/backoffice/templates/operational-dashboard.html
```diff
@@ -1,113 +1,95 @@
...
-
-{% end %}
-
-{% block content %}
-<div class="container operational-dash">
-    <div class="row text-center">
-    <h1 class='text-center'>{{ _("Operational dashboard") }}</h1>
-    </div>
+<!-- Dashboard filter navbar -->
+{% block dashboard_filter %}
```
rth, line 4: This diff is a bit hard to review, but essentially I split `operational-dashboard.html` into two templates: one base template with the general page layout and this file with the HTML code of each block.

So far the functionality is not changed, but this will help putting these different blocks into tabs.

rth:
Also pushed a copy of this to an upstream branch https://github.com/icubam/icubam/tree/dashboard-plots-v2 to make trying it out easier.

icubam/predicu/plot/__init__.py
```diff
@@ -122,29 +122,28 @@ def plot(
...
-      figure=fig,
-      **tikzplotlib_kwargs,
-      standalone=True,
-    )
-  elif output_type in ["png", "pdf"]:
-    fig.savefig(os.path.join(output_dir, f"{plot_name}.{output_type}"))
   else:
-    raise ValueError(f"Unknown output type: {output_type}")
-  plt.close(fig)
+    figs, tikzplotlib_kwargs = plot_fun(data=cached_data.copy())
```
tgy, line 129: why pass data to the function it it does not ask for it?

rth: Thanks! True, actually a more serious issue is that copy doesn't deepcopy the dataframes. So I should iterate over required keys manually and copy them.

Issue #351, opened 2020-04-20
docker builds currently failing


rprimet:
See for instance https://github.com/icubam/icubam/runs/599546914?check_suite_focus=true

Extract from the logs

```
/home/icubam/icubam/predicu/tests/test_plot.py:51
_ ERROR at setup of test_integration_generate_plots[barplot_healed_per_dept_per_day] _
file /home/icubam/icubam/predicu/tests/test_plot.py, line 51
  @pytest.mark.integration
  @pytest.mark.parametrize("name", PLOTS)
  def test_integration_generate_plots(name, integration_config, tmpdir):
E       fixture 'integration_config' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

rth:
You need to copy the `conftest.py` file inside docker for pytest to work with fixtures defined there.

rth:
It's surprising we didn't catch this when merging https://github.com/icubam/icubam/pull/314

rth:
Closing as resolved in https://github.com/icubam/icubam/pull/352

PR #352, opened 2020-04-20
Fix missing config copy in Dockerfile

pgraverdy:
Solves #351 

Fix error in Action for docker image creation due to missing conftest.py.

The conftest.py file has been added in Dockerfile, which uses explicit copy of specific files and folders, and not a blanket copy of everything in the build folder except those in .dockerignore. 
While this prevents building a docker image with extra/unwanted files added in the local build folder (on test/dev machines) it also has the potential to create bugs if the Dockerfile has not been properly updated.

@rth @marccarre any recommendation for this ? (do blanket copy anyway as (1) on CICD, no risks or unwanted files in the image and (2) the images built on dev machines are just local anyway).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,9 +2,9 @@
 .github
 .gitignore
 data
-deploy_configs
-deploy_configs.tgz
 icubam.egg-info
 docker/configs/certbot
 docker/scripts
 docker/resources
+test.db
+icubam.db
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,5 +6,5 @@ icubam.egg-info
 docker/configs/certbot
 docker/scripts
 docker/resources
-test.db
-icubam.db
+**/*.db
+
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,7 +36,7 @@ COPY scripts scripts
 RUN  mkdir resources
 
 # copy specific files on root folder
-COPY LICENSE pyproject.toml README.md requirements.txt requirements-test.txt setup.py ./
+COPY LICENSE pyproject.toml README.md requirements.txt requirements-test.txt setup.py conftest.py ./
 
 # pip/setup tools requires version inside .git
 COPY .git ./.git


rth:
> The conftest.py file has been added in Dockerfile, which uses explicit copy of specific files and folders, and not a blanket copy of everything in the build folder except those in .dockerignore.

I'm not sure I understand. Is this
https://github.com/icubam/icubam/blob/16fb298269d0edd045a03bdcc002e77ba8c90662/conftest.py#L35
the problem? Or just the existence of `conftest.py`? By default that fixture is not used unless you run integration tests: https://github.com/icubam/icubam/blob/master/doc/contributing.md#integration-tests The file needs to be run, but the corresponding fixures are not going to be run unless you pass the `--icubam-config` to pytest.

rth:
Shouldn't some tests fail if tests cannot be run inside Docker though? Do you understand why CI is green on master now?

pgraverdy:
Indeed, the issue was the missing conftest.py in the Dockerfile (was not copied), hence when running the tests inside a container, it was failing. Adding the file inside the Dockerfile has solved the issue AFAICT.

rth:
> Adding the file inside the Dockerfile has solved the issue AFAICT.

OK, great. I think copying files one by one is still fine, because there could be potentially sensitive files in the project dir, particularly when building locally. Ideally if CI fails, it should be easy enough to detect that some file is missing.

.dockerignore
```diff
@@ -2,9 +2,9 @@
...
-deploy_configs
-deploy_configs.tgz
 icubam.egg-info
 docker/configs/certbot
 docker/scripts
-docker/resources
\ No newline at end of file
+docker/resources
+test.db
+icubam.db
```
rth, line 10: Maybe more generally, `*.db` unless we can't use whildcards in `.dockerignore` ?

pgraverdy: Indeed. replaced with globstar.

rth:
Thanks @pgraverdy !

Issue #339, opened 2020-04-20
Deprecate --mode flag in config


rth:
Following discussion between @dulacarnold and @olivierteboul we can deprecate the `mode` flag in config in favor of just using different config files.

Issue #358, opened 2020-04-20
Docker deployment design modifications


dulacarnold:
To facilitate deployment of new instances on GCP, what are your thoughts on the following changes in docker approach:
* Make it so that by default the docker image mounts an external folder for all stateful elements: db, configs, env file, upload dir, image plots (I think that is all?)
* Either deprecate the --mode flag or default it to prod.
* Concentrate on the non-nginx image as our main deployment tool (on cloud this looks like the right way to go).
* Add a script that does a regular dump of the sqlite db to a .dump file in the same directory.  This is so that snapshots that are taken don’t happen during an insert and create a corrupt file.
* Make sure that on first run, everything boots and it creates the necessary db file.  We can also provide default config/env files that will get copied into the mounted directory.  We should also make sure all directories such as ./upload and plot directories get create if not present.
* Make restarting the services / upgrading versions easy  (potentially automatic, with this deployment scenario there are no more secrets as part of the docker deployment step, secrets can either be put in place by hand or deployed by a seperate CI mechanism).

rprimet:
+1 for the folder but I'd mount two (one readonly for the config, one RW for the DB etc)

PR #360, opened 2020-04-21
Deprecate "mode" option

rth:
As discussed in https://github.com/icubam/icubam/issues/358 deprecates the `--mode` option in favor of using separate config files for `dev` and `prod`.

Closes https://github.com/icubam/icubam/issues/339

This should address it in code and scripts, but not in docker setup. There is a fair amount of changes required there, so someone should feel free to continue this here or in a separate PR.

Just make sure that the squashed commit message contains the [`Co-authored-by:`](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-github`) tag of all participants for attribution.


dulacarnold:
Great work, this was indeed everywhere!

pgraverdy:
@dulacarnold  @rth I updated the docker side to reflect the change (transmitting the ENV_MODE environment variable from the build environment to the docker container).

However, as there are now different names for the config.toml files depending on the environment, I cannot just mount the file as is currently done. I must update the docker mount operations so that a folder is mounted (with all the resources in it). This will be cleaner and more flexible. What do you think?

dulacarnold:
> @dulacarnold @rth I updated the docker side to reflect the change (transmitting the ENV_MODE environment variable from the build environment to the docker container).
> 
> However, as there are now different names for the config.toml files depending on the environment, I cannot just mount the file as is currently done. I must update the docker mount operations so that a folder is mounted (with all the resources in it). This will be cleaner and more flexible. What do you think?

Agreed, I think we should consider that the 'default' file is always icubam.toml (I know this PR currently has an icubam_dev.toml).  And then if someone wants to have different settings locally, up to them to make different files.  @rth would that conflict with what you had in mind?  

@pgraverdy this doesn't go against mounting a full folder, as that would  effectively become the 'resources' folder (or whatever we want to call it), which would also have icubam.env, icubam.db and whatever else.  So let's do that anyways, but try and just consider that the config file has one default name: icubam.toml, and a different version will be provided when in production.

docker/build_fake_dbs.sh
```diff
@@ -9,10 +9,10 @@ conda activate icubam
 pwd
 ls -a
 echo "Build test db"
-PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config.toml --mode=dev
+PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config_dev.toml
```
dulacarnold, line 12: To go along with @pgraverdy's comment, I feel like we should just have a single config file.  The prod/dev/test whatever approach is not something I've seen in many other software packages, and can be easily dealt with by simply allowing for /whatever/ config file you would want.

docker/start_server.sh
```diff
@@ -2,4 +2,4 @@
 
 # launch the server
 echo "launch icubam server in mode ${ENV_MODE}"
-python3 scripts/run_server.py --port 8888 --mode="${ENV_MODE}" --config=/home/icubam/resources/config.toml
+python3 scripts/run_server.py --port 8888 --config="/home/icubam/resources/config_${ENV_MODE}.toml"
```
dulacarnold, line 5: So here I would just consider there to only be a single config.toml, no need to even use the ENV_MODE variable.  I know that for myself I just run everything with prod all the time, and depending on where I am I change the config files, or point to a different config file if needed.

rth:
> Agreed, I think we should consider that the 'default' file is always icubam.toml (I know this PR currently has an icubam_dev.toml)

Sounds good. I added `icubam_dev.toml` because I wasn't sure how to adapt the script, but a single config file for Docker sounds good. As long as we have a config file that allows to run servers locally (and one for testing as currently there).

pgraverdy:
So to sum up:
- remove the ENV_MODE variable everywhere
- in the compose files, mount the "./resources" folder in the containers (default value of ICUBAM_RESOURCE_PATH, can be another folder)
- assume there is a icubam.toml folder in it that is used.

Thus
- when launching directly with python, we use the resource folder already there
- when deploying in production, we change the ICUBAM_RESOURCE_PATH envvar to the path of the production config/db folder.

rth:
> One question regarding the toml file for unit tests (and the related testing files). Should we have them in a different folder (e.g., "test_resources" with the test.toml renamed as icubam.toml)

I think test data use from unit tests (including test config files) should be part of the ICUBAM package (i.e. inside the python icubam folder) https://github.com/icubam/icubam/issues/296#issuecomment-616095233 Otherwise all tests break as son as one tries to change the current directory.

No one actually starts the server with `resources/test.toml` as far as I understand?

If it's a blocker here, I can make a PR.

docker/build_fake_dbs.sh
rth, line 12:
```suggestion
PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config.toml
```

docker/build_fake_dbs.sh
```diff
@@ -9,10 +9,10 @@ conda activate icubam
...
 
 echo "Build icubam db"
-PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config.toml --mode=prod
+PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config_prod.toml
```
rth, line 15:
```suggestion
PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config.toml
```

docker/configs/nginx/app.conf.prod
```diff
@@ -3,11 +3,11 @@
 #
 
 upstream app-server {
-        server icubam-server:8888;
+        server icubam_server:8888;
```
rth, line 6: Not really related to this PR, but it might have been more readable to define it as `icubam_www_server` to deflect its name in the code.

icubam/config.py
```diff
@@ -8,7 +8,7 @@
 import toml
 
 DEFAULT_DOTENV_PATH = "resources/icubam.env"
-DEFAULT_CONFIG_PATH = "resources/config.toml"
+DEFAULT_CONFIG_PATH = "resources/config_dev.toml"
```
rth, line 11:
```suggestion
DEFAULT_CONFIG_PATH = "resources/config.toml"
```

docker/build_fake_dbs.sh
pgraverdy: Thanks. indeed I noticed and changed already on my side, fixing the different shell scripts for starting docker currently

docker/build_fake_dbs.sh
pgraverdy: Thanks. indeed I noticed and changed already on my side, fixing the different shell scripts for starting docker currently

docker/configs/nginx/app.conf.prod
pgraverdy: Good point. I'll do the change. Thanks

pgraverdy:
@rth can you check/validate my changes ? (cannot add you as a reviewer)

@dulacarnold can you review these changes (and put the `Co-authored-by: tag` of all participants for attribution).

doc/docker.md
```diff
@@ -202,30 +194,36 @@ started.
...
 Remove/force the icubam containers
 ```
-docker rm -f icubam-server icubam-sms-server icubam-bo-server
+docker rm -f icubam_www_server icubam_sms_server icubam_bo_server
 ```
 
 Get the logs of the server
 ```
-docker logs icubam-server
+docker logs icubam_www_server
```
dulacarnold, line 202: Ok, and I guess this also works fine for the other instances as well? (their logs are also useful)

pgraverdy: Yes. this is a standard docker command.

icubam/messaging/email_sender_test.py
```diff
@@ -19,7 +19,7 @@ def setUp(self):
     os.environ['SMTP_USER'] = SMTP_USER
     os.environ['SMTP_PASSWORD'] = SMTP_PASSWORD
     os.environ['EMAIL_FROM'] = EMAIL_FROM
-    self.config = config.Config('resources/test.toml', mode='dev')
+    self.config = config.Config('resources/test.toml')
```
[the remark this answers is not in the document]
rth: No, but this is the specific file to run unittest. You cannot/shouldn't run unitttests on the production config.

PR #362, opened 2020-04-21
Plot generation script

rth:
Adds a script to manually generate plots for the dashboard https://github.com/icubam/icubam/issues/284 instead of the callback https://github.com/icubam/icubam/pull/222

There is some redundancy with `icubam.predicu.plot.__main__` but I thin we should move that one, one level up and expose it as `python -m icubam.predicu plot`, otherwise if each module has a script it's a bit hard to discover.

### DIFF
--- a/icubam/predicu/__main__.py
+++ b/icubam/predicu/__main__.py
@@ -1,13 +1,37 @@
-import argparse
+from pathlib import Path
 import datetime
 import logging
 import os
+import sys
 
+import asyncio
+import click
+
+from icubam.config import Config
+from icubam.db.store import create_store_factory_for_sqlite_db
 from icubam.predicu.data import load_bedcounts
+from icubam.predicu.analytics_server import AnalyticsCallback
 
 
-def export_data(args):
-  output_dir = args.output_dir
+@click.group()
+def cli():
+  pass
+
+
+@cli.command(name="export")
+@click.option("--output-dir", "-o", default="/tmp", type=str)
+@click.option("--api-key", default=None)
+@click.option("--spread-cum-jump-correction", default=False, type=bool)
+@click.option("--icubam-host", default="localhost", type=str)
+@click.option(
+  "--max-date",
+  default=None,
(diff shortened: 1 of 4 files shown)


icubam/predicu/__main__.py
```diff
@@ -19,31 +51,52 @@ def export_data(args):
...
+  if not output_dir.exists():
+    if not output_dir.parent.exists():
+      click.echo(f"Output directory {output_dir}!", err=True)
+      sys.exit(1)
+    # otherwise necessary folder will be created by generate_plots
+
+  store_factory = create_store_factory_for_sqlite_db(config)
+  callback = AnalyticsCallback(output_dir, store_factory)
+  eventloop = asyncio.new_event_loop()
+  eventloop.run_until_complete(callback.generate_plots())
```
rth, line 96: It's a bit over-engineered but I just want to re-use the same callback as the one used to generate the plots with the backoffice server.

icubam/predicu/tests/test_export_data.py
```diff
@@ -18,17 +16,11 @@ def test_export_data(tmpdir, monkeypatch):
...
-  TestArgs = collections.namedtuple(
-    'TestArgs', [
-      'output_dir', 'api_key', 'max_date', 'icubam_host',
-      'spread_cum_jump_correction'
-    ]
-  )
-  test_args = TestArgs(
+  test_args = dict(
     output_dir=str(tmpdir),
     api_key='test_api_key',
```
dulacarnold, line 21: Is this thing hitting the /db/* API with a key to be able to get at the data?  How does it know what key?

rth: No I'm not using this CLI script. I only migrated it from argparse to click so it can be used as part of the same CLI group.

The added plotting scipt directly queries the DB in https://github.com/icubam/icubam/blob/9cf8f0d02755a231769abba9f6907a4781306f8b/icubam/predicu/analytics_server.py#L18

PR #363, opened 2020-04-21
add sqlite3 to be able to backup db from a docker container

VincentRouvreau:


rth:
How can it run DB without sqlite3 though? Does it mean that there are 2 versions of sqlite in that image now? Or is it just the CLI package? Just wondering.

rprimet:
@rth I'm not positive but it may be the CLI package only, and the python install uses `libsqlite3`.
(at least running `which sqlite3` from within the container yields no result)

dulacarnold:
Do we want to force a specific version here?  Will it be able to dump upserted files?

rth:
Yes, you are right, it's only the CLI,
```
root@267d7ff5b9cd:/# apt show sqlite3 
Package: sqlite3
Version: 3.27.2-3
[...]
APT-Sources: http://deb.debian.org/debian buster/main amd64 Packages
Description: Command line interface for SQLite 3
```
the version is OK as well (3.24).

rprimet:
@dulacarnold I think we base our image on debian buster, right? 
https://packages.debian.org/stable/sqlite3
So a system install gets us sqlite 3.27 which is good for our needs I think?

PR #365, opened 2020-04-21
Add restrict_to_region to public data loading fun

tgy:
+ other things... @rth to the rescue


icubam/predicu/data.py
```diff
@@ -1,12 +1,11 @@
-from typing import Optional
-
 import inspect
 import itertools
 import json
 import logging
 import pickle
 import urllib.request
 from pathlib import Path
+from typing import Optional
```
rth, line 8: The previous ordering was actually isort I think...

icubam/predicu/data.py
```diff
@@ -86,7 +85,11 @@ def load_if_not_cached(data_source, cached_data, **kwargs):
     )
   if cached_data is None or data_source not in cached_data:
     load_data_fun_signature = inspect.signature(load_data_fun)
-    kwargs = {}
+    kwargs = {
+      k: v
+      for k, v in kwargs.items()
+      if k in load_data_fun_signature.parameters
+    }
```
rth, line 92: Sorry it was me removing this. Why is it necessary. Functions will be called with defaults values anyway?

tgy: `api_key` and `icubam_host` were not propagatade to `load_combined_bedcounts_public`

rth:
Merging. One CI failure is because Github Packages are down.

PR #366, opened 2020-04-21
Filter plots by region in the dashboard

rth:
Still related to #284


rth:
I thinks this is ready for review. The overall solution of mapping figure URLs to plot filenames is a mess (in particular because plotting module and the BD use completely different metrics), so it's clearly only a temporary workaround. At the same time this also include two relevant fixes,
 - only show regions with some data in the dropdown button
 - keep the last open tab in localstorage and re-open it when page is reloaded (e.g. a different region is selected).

Currently this aims to show graphs for all regions when that option is selected and otherwise show 1 plot for the selected region in the Covid+ tab.

PR #367, opened 2020-04-21
Dashboard available from outside with API key

olivierteboul:


icubam/predicu/plot/operational_dashboard.py
```diff
@@ -0,0 +1,204 @@
+
```
rth, line 1: This should be put into a different folder, maybe ` icubam/predicu/dashboard`. All modules in ` icubam/predicu/plot/` are assumed to be plots with a specific API (e.g. have a plot function that returns something) and are automatically tested by pytest.

dulacarnold: Ah good point!

rth:
Nice work! Just one comment above otherwise LGTM (after a superficial review). 

 pre-commit (https://github.com/icubam/icubam/blob/master/doc/contributing.md#code-style) does help for dealing with linters :)

icubam/predicu/plot/operational_dashboard.py
olivierteboul: cool! done

olivierteboul:
> Nice work! Just one comment above otherwise LGTM (after a superficial review).
> 
> pre-commit (https://github.com/icubam/icubam/blob/master/doc/contributing.md#code-style) does help for dealing with linters :)

thanks! Gabriel also pointed it to me. better indeed

PR #369, opened 2020-04-22
Move icubam.predicu.plot CLI one level up

rth:
Moves the CLI from,
```
python -m icubam.predicu.plot ..
```
to
```
python -m icubam.predicu plot ..
```
and uses click for argument parsing. The only other difference is that instead of the optional `--plots [..]` argument, it's now an optional positional argument at the end (as click doesn't support nargs="*").

Also adds the CLI documentation.


icubam/predicu/plot/__init__.py
```diff
@@ -98,10 +98,7 @@ def plot_int(
 
 PLOTS = []
 for path in os.listdir(os.path.dirname(__file__)):
-  if path.endswith(".py") and any(
-    path.startswith(prefix)
-    for prefix in ["barplot", "lineplot", "scatterplot", "stackplot", 'flux']
-  ):
+  if path.endswith(".py") and path != "__init__.py":
```
rth, line 101: Just test any file in that folder. This will ensure that unit tests are run on any new file in that folder.

dulacarnold: So pretty quickly we many want to only generate certain plots, is that easy to do?

rth: Yes, this only affects tests. For the CLI or the `generate_plots` function you still have a `plots` argument that one can either leave to None to generate all plots or provide the list of specific plots to generate. This is for instance what we are doing in https://github.com/icubam/icubam/blob/fdc51762e880a50a21dec73fd7e2d46ad9135e35/icubam/predicu/analytics_server.py#L23
to only generate `flux_dept_lits_dept_visu_4panels` plots.

Issue #286, opened 2020-04-16
refactor preprocessed data loading


tgy:
currently, the already-loaded data is preprocessed using `load_bedcounts` by passing it as `cache_data = {'raw_icubam': data}`. tihs is not intuitive and the preprocessing v.s. loading architecture should be revised

Issue #299, opened 2020-04-16
predicu integration


rth:
To better integrate predicu into icubam, following ideas were discussed,
 - [ ] rename the `icubam.predicu` module to `icubam.analytics`
 - [ ] make required data files a bit more language independent https://github.com/icubam/icubam/issues/296 
 - [ ] In `test_generate_plots` reconsider using a pre-generated `tests/data/bedcounts.csv` with fake data. For instance, for me this test passes, but the `generate_plots` fails when run on the pre-procution DB in https://github.com/icubam/icubam/pull/222 with,
```
    fig, tikzplotlib_kwargs = plot_fun(data=cached_data[data_source].copy())
  File "/home/rth/src/icubam/icubam/predicu/plot/barplot_deaths_per_day.py", line 11, in plot
    d = d.groupby(["date", "department"]).sum().reset_index()
  File "/home/rth/miniconda3/envs/icubam/lib/python3.8/site-packages/pandas/core/frame.py", line 5801, in groupby
    return groupby_generic.DataFrameGroupBy(
  File "/home/rth/miniconda3/envs/icubam/lib/python3.8/site-packages/pandas/core/groupby/groupby.py", line 402, in __init__
    grouper, exclusions, obj = get_grouper(
  File "/home/rth/miniconda3/envs/icubam/lib/python3.8/site-packages/pandas/core/groupby/grouper.py", line 598, in get_grouper
    raise KeyError(gpr)
KeyError: 'date'
```

cc @tgy

rth:
Closing as resolved since https://github.com/icubam/icubam/pull/389

Issue #285, opened 2020-04-16
speed up load_bedcounts's preprocessing


tgy:
2 slow functions right now:
- `enforce_daily_values_for_all_icus`
- `aggregate_multiple_inputs`

there are some obvious ways to replace for loops with pandas operations.

just have to make sure nothing is broken (result should be exactly the same before and after the optimisation changes and tests should be written)

PR #371, opened 2020-04-22
Refactor predicu.data

rth:
This does an in depth refactoring of `icubam.predicu.data` with respect to caching and when the preprocessing is applied. This is necessary to consistently apply preprocessing independently of how data is obtained (e.g. in https://github.com/icubam/icubam/issues/284)

Closes https://github.com/icubam/icubam/issues/286, related to https://github.com/icubam/icubam/issues/299

Currently preprocessing is part of the `load_*` functions, and is tightly linked to when data is loaded from HTTP API. So if data is loaded directly from DB it's difficult to apply the same pre-processing.

This PR,
 - moves preprocessing fuctions to separate module `predicu.preprocessing`
 - `load_*` functions only load the raw data from HTTP API and nothing else
 - preprocessing and filtering of data is done in `preprocess_bedcounts` function. Other preprocessing functions could similarly be later added and called via the `preprocess_data(data_source, data, **kwargs)`.
 - removed `load_icubam` because it's the same thing as `load_bedcounts`, but with each was previously applying additional pre-processing.
 - all assembling of required data and preprocessing is applied in `icubam/predicu/plot/__init__.py:plot` to call plotting functions with the data they need
 - removed `load_combined_bedcount_public` because it's unclear how to make it work with the "loaders don't do preprocessing" idea. Instead the `combined_bedcount_public` is assembled for plots that need it in `plot/__init__.py:plot`
 - caching raw data is now only relevant to plotting functions that prepare data and don't need to be propagated to all loader functions.
 - remove `normalize_colum_names` function as no longer necessary. The fact that we needed it was a symptom of the issue.

This will not be easy to review, and we need to ensure all generated plots are the same, bit it should help for the maintainability of data loaders.

TODO:
 - [x] fix failing tests
 - [x] make sure generated figures are the same
 - [x] make sure exported data is the same via the DB HTTP API.


icubam/predicu/tests/data/bedcounts.csv
```diff
@@ -1,5 +1,5 @@
 # This file contains a randomly generated dataset for testing
-icu_name,date,datetime,department,n_covid_deaths,n_covid_healed,n_covid_transfered,n_covid_refused,n_covid_free,n_ncovid_free,n_covid_occ,n_ncovid_occ,region
+icu_name,create_date,datetime,icu_dept,n_covid_deaths,n_covid_healed,n_covid_transfered,n_covid_refused,n_covid_free,n_ncovid_free,n_covid_occ,n_ncovid_occ,icu_region_id
```
rth, line 2: This now becomes the raw data before preprocessing

icubam/www/handlers/db.py
```diff
@@ -87,13 +87,7 @@ def get(self, collection):
...
       data = store.to_pandas(get_fn(), max_depth=1)
       if collection == 'all_bedcounts' and should_preprocess:
-        # this cached_data dict is just a way to tell predicu not to load the
-        # data from ICUBAM and use this already loaded data instead
-        cached_data = {'raw_icubam': data}
-        data = predicu.data.load_bedcounts(
-          cached_data=cached_data,
-          preprocess=True,
-        )
+        data = preprocess_bedcounts(data)
```
rth, line 90: Separating processors and loaders avoids doing awkward things with cached_data.

icubam/predicu/plot/barplot_inputs_per_day.py
```diff
@@ -1,7 +1,7 @@
 import matplotlib.pyplot as plt
 import seaborn as sns
 
-data_source = ["icubam"]
+data_source = ["bedcounts"]
```
rth, line 4: `icubam` should contain the same data (but possibly with different pre-processing and bedcounts)

icubam/predicu/plot/__init__.py
```diff
@@ -116,28 +120,57 @@ def plot(
...
+
+    plot_data[name] = preprocess_data(name, data, **preprocess_kwargs)
+
+  if needs_combined_data:
+    plot_data["combined_bedcounts_public"] = combine_bedcounts_public(
+      plot_data['public'], plot_data['bedcounts']
     )
+    data_source.remove("bedcounts")
+    data_source.remove("public")
+    data_source.append("combined_bedcounts_public")
```
rth, line 165: There is quite a bit of logic in this function to fetch the data that is necessary for plots. Not sure if there is a better way.

rth:
> icubam [data source] should contain the same data (but possibly with different pre-processing and bedcounts)

OK it was a bit more complicated than that. "icubam" was the same datasource than "bedcounts" but with different preprocessing, now corresponding to `preprocess_bedcounts(..., full=False)`

So here I added a `preprocesing_args = {'bedcounts': {'full': False}}` variable to the plot script to pass optional preprocesing arguments to the function that prepares data.

I have checked than,
 - the figures in the IUCBAM dashboard now look OK (so the correct pre-processing is applied)
 - figures generated with `python -m icubam.predicu plot` produce indentical plots to those on master. I got an error https://github.com/icubam/icubam/issues/372 for 2 plots, but I also see it in master.
 - querying `db/all_bedcounts?format=csv&API_KEY=...&preprocess=True` produces the same output in this PR and on master.

So overall I'm fairly confident this shoudn't break anything. Feedback weclome.

rth:
BTW, this makes generation of all plots in one go a bit slower because now pre-processing is applied once per plot (since pre-processing options can be different). We'll just have to optimize two functions from https://github.com/icubam/icubam/issues/285 to make it faster.

icubam/predicu/plot/__init__.py
tgy: maybe create a fonction `get_raw_sources_needed_for_data_source(data_source : str)` such that
```python
raw_sources = get_raw_sources_needed_for_data_source("combined_bedcounts_public")
assert raw_sources == {"bedcounts", "public"}`
```
maybe that would make the code clearer

icubam/www/handlers/db.py
tgy: 👍

icubam/predicu/plot/__init__.py
rth: Thanks a lot for the review! Yes, I'll try to do it in a follow up PR, as this one has already way too many changes :)

PR #374, opened 2020-04-22
predicu: faster pre-processing

rth:
Closes https://github.com/icubam/icubam/issues/285 (well other optimizations are also possible).

Make `preprocess_bedcounts` run on all data ~55% faster: 7.1s on master, 3s in this PR.

This should also make plotting tests faster (as well as data export with pre-processing)

TODO
 - [ ] check that predicu plots hasn't changed.


icubam/predicu/preprocessing.py
```diff
@@ -238,8 +223,7 @@ def enforce_daily_values_for_all_icus(d):
               ).first().reset_index()[["icu_name", "region"
                                        ]].itertuples(name=None, index=False)
   )
-  for date, icu_name in itertools.product(dates, icu_names):
-    sd = d.loc[(d.date == date) & (d.icu_name == icu_name)]
+  for (date, icu_name), sd in d.groupby(['date', 'icu_name']):
```
rth, line 226: By default groupby sorts group keys already https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html

tgy: careful here. the reason i was doing `itertools.product` was to ensure that there is a value for every icu_name and every date.

let `N` be the number of dates and `K` be the number of ICUs.

the goal of the function is to ensure `N x K` rows in the data. this means that even if we don't have data for an ICU at a specific date, we will still have a row for that date and ICU in the preprocessed data.

if you do a groupby instead, you're not ensuring that.

tgy: here a better way to do this would be to use a `pd.date_range` over the entire period, reindex the dataframe of each icu with the date range and input the missing values with some `ffill` or `bfill`

rth: Very good point, thanks! I think we can still do a groupby but with observed=False and using categorical dtype which would do this carthesian product.

tgy: great, did not know that existed. you seem to know all the magic pandas tricks 🐼

rth: Yes, reindexing + imputation totaly makes sense.

rth:
OK, so I rewrote `enforce_daily_values_for_all_icus` to use re-indexing on full time range. Overall the results are the same (and the processing is still ~60% faster), however occasionally the one of columns will be different by a few counts with respect to the preprocessing on master (on the production data). Still investigating what this could be due to, I think it likely related to how we handle submissions with multiple entries per day.

rth:
Yes, the issue was with tie breaking when sorting by date instead of datetime. 
The results are now identical to master.

PR #375, opened 2020-04-22
Telegram support

olivierteboul:
- Add function to send and fetch from telegram.
- Add function to identify user from a start button
- Add a webhook to directly receives updates from telegram
- Adds a TelegramSetup class that is responsible for deciding if and how to use telegram support.
- Add a sender class that decides through which channel (sms, email, telegram) to send a message.
- Add integration in message server
- Added the a class for format and translate the messages before sending them.
- Added a link to the telegram bot in update form in case telegram is set and the user has not registered yet.
- Unit tests for every new class

Needs:
 - `python3 -m scripts.migrate_db` to migrate the db with the new column
 - #378 for proper tokens to be sent to the user


resources/config.toml
```diff
@@ -14,13 +14,21 @@ default_locale='fr-fr'
...
-  base_url = "http://localhost:8888/"
+  [server.dev]
+  PORT = 8887  # will be lower cased when reading.
+  base_url = "http://localhost:8887/"
   upload_dir = "uploads/"
   consent = 'resources/consent.html'
   num_days_for_stale = 1.0
 
 [messaging]
+  [messaging.dev]
```
rth, line 25: Maybe it would be worth syncing with master as this is still using the dev/prod config version.

rth:
The API rate limit [seem to be around 30 sent messages/s](https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this) so it would likely be necessary to spread out Telegram actions. Is this what `telegram_updates_every` does already?

olivierteboul:
> The API rate limit [seem to be around 30 sent messages/s](https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this) so it would likely be necessary to spread out Telegram actions. Is this what `telegram_updates_every` does already?

I believe 30 messages / s is fine for now, but it's good to know.

Basically there are two ways to get the messages from telegram to our server. Either set up a webhook (best way imo), so that telegram sends us the messages (updates) on our endpoint. Or we can periodically call telegram to ask for the updates.

In this PR I am implementing the second methods because it is either in dev mode to do. To setup a webhook we need a valid domain, which we have in prod but not on our local machines. (ngrok can be convenient here). The `telegram_updates_every` exactly the time to wait beetween asking again for the update to the telegram servers.

resources/config.toml
olivierteboul: my bad. i ll fix it

resources/test.toml
```diff
@@ -15,7 +15,7 @@ sqlite_path = ":memory:"
 
 [server]
   PORT = 8888  # will be lower cased when reading.
-  base_url = "https://localhost"
+  base_url = "http://localhost"
```
rth, line 18: Why do we need https? It will not work for the test server anyway without certificates?

resources/config.toml
```diff
@@ -26,6 +26,12 @@ default_locale='fr-fr'
   PORT = 8889  # will be lower cased when reading.
   base_url = "http://localhost:8889/"
   timeout = 2
+  telegram_bot = "test_test_icu_bot"
+  telegram_updates_every = 10
+  [messaging.prod]
+  PORT = 8889  # will be lower cased when reading.
+  base_url = "http://localhost:8889/"
+  timeout = 2
```
rth, line 34: The above 3 lines can probably be removed? Since we no longer use dev/prod setup?

icubam/messaging/sender.py
```diff
@@ -0,0 +1,57 @@
...
+    """Sends the message to a single user."""
+    if self.telegram_bot is not None and user.telegram_chat_id is not None:
+      await self.telegram_bot.send(user.telegram_chat_id, msg.html)
+      return True
+    elif user.email is not None and self.email_sender is not None:
+      self.email_sender.send(user.email, 'ICUBAM', msg.html)
+      return True
+    elif self.sms_sender is not None and user.telephone is not None:
+      self.sms_sender.send(msg.phone, msg.text)
+      return True
```
rth, line 56: Maybe instead of True / False, return the string 'sms', 'email', 'telegram_message' or None?

Currently tests, assert that `await ..send()` is True after changing some config but it's actually hard to know whether the right sender was used.

icubam/messaging/telegram/integrator.py
```diff
@@ -0,0 +1,63 @@
...
+      self.bot = bot.TelegramBot(config) if tg_bot is None else tg_bot
+      self.processor = updater.UpdateProcessor(
+        config, db, self.queue, scheduler, tg_bot=self.bot
+      )
+      self.fetcher = updater.UpdateFetcher(config, self.queue, tg_bot=self.bot)
+
+  @property
+  def uses_webhook(self):
+    url = self.config.server.base_url
+    return 'https' in url and 'localhost' not in url
```
rth, line 23: Could you add an inline comment explaining this? It could have been 127.0.0.1 as well FWIW.

icubam/messaging/telegram/integrator.py
```diff
@@ -0,0 +1,63 @@
+import tornado.queues
+from icubam.messaging.telegram import bot, updater, webhook
+
+
+class TelegramSetup:
+  """Sets up the telegram in the server support depending on the config."""
```
rth, line 6: Maybe you could link the the relevant Telegram API docs in the docsting? In particular I'm not sure I understand the conditions under which the webhook is used.

icubam/messaging/telegram/updater.py
```diff
@@ -0,0 +1,101 @@
...
+    self.queue = queue
+    self.scheduler = scheduler
+    self.bot = bot.TelegramBot(config) if tg_bot is None else tg_bot
+    # TODO(olivier): replace with authenticator.
+    self.token_encoder = token.TokenEncoder(self.config)
+    admins = self.db.get_admins()
+    if admins:
+      self.admin_id = admins[0].user_id
+    else:
+      self.admin_id = self.db.add_default_admin()
```
rth, line 53: Could you please add an inline comment on why we need these admin_id when updating users?

icubam/www/token.py
```diff
@@ -22,3 +23,36 @@ def decode(self, token: str):
 
   def encode_data(self, user, icu) -> str:
     return self.encode((user.user_id, icu.icu_id))
+
+  def authenticate(self, token: str, db) -> Tuple:
```
rth, line 27: So this is identical to `decode_token` previously, right?

resources/config.toml
rth: Above comment is still relevant.

icubam/messaging/sender.py
```diff
@@ -0,0 +1,57 @@
...
+      logging.warning(f"Cannot set email_sender {e}")
+      self.email_sender = None
+
+    # This might return be None if telegram is not properly set.
+    self.telegram_bot = integrator.TelegramSetup(config, db, tg_bot=tg_bot).bot
+
+  async def process(self):
+    """Keeps reading the sending queue and does send the messages."""
+    async for msg in self.queue:
+      try:
```
rth, line 38: This function is never called in tests.

icubam/messaging/sender.py
```diff
@@ -0,0 +1,57 @@
...
+
+  async def send(self, msg, user) -> bool:
+    """Sends the message to a single user."""
+    if self.telegram_bot is not None and user.telegram_chat_id is not None:
+      await self.telegram_bot.send(user.telegram_chat_id, msg.html)
+      return True
+    elif user.email is not None and self.email_sender is not None:
+      self.email_sender.send(user.email, 'ICUBAM', msg.html)
+      return True
+    elif self.sms_sender is not None and user.telephone is not None:
```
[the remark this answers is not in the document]
olivierteboul: there is some decision to be taken here: 1) the order 2) can we allow several channels. I have no preference, i have put sms at the end since it s the most expensive.

icubam/messaging/sender.py
olivierteboul: true, also harder to test. i ll change to something more meaningful.

icubam/messaging/telegram/integrator.py
olivierteboul: good one. The telegram webhook can only be set on a https route. when you are in dev mode on your machine you usually dont have this.

icubam/messaging/telegram/integrator.py
olivierteboul: Sure. will do.
Basically you have two ways to get the messages from telegram to our server:
1) periodically ask telegram to get the last messages.
2) set a webhook so that telegram forwards directly every message (call update) to us when they receive it. The webhook must be a https url.

The integrator decides if we are getting the messages via webhook or periodic callback and sets the right stuff in the message server. in case telegram is not supported by our config, it just does nothing (say you have no bot set up).

icubam/messaging/telegram/updater.py
olivierteboul: sure. though this is how the store is designed.

icubam/www/token.py
olivierteboul: absolutely!

icubam/messaging/sender.py
olivierteboul: yeah... i am not sure yet how to unit test it, but i ll give it a try.

olivierteboul:
Thanks for the comments Roman! I ll update the code soon.

icubam/messaging/sender.py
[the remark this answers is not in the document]
rth: Telegram first sounds good.

icubam/messaging/message_formatter.py
```diff
@@ -0,0 +1,35 @@
...
+  """This class aims at formatting the messages sent to the user.
+
+  Formatting means turning the data in the message into a localized text or
+  html.
+  """
+  ROOT_PATH = os.path.abspath(__file__)
+  for _ in range(3):
+    ROOT_PATH = os.path.dirname(ROOT_PATH)
+
+  ROOT_PATH = os.path.join(ROOT_PATH, 'resources/messages/')
```
rth, line 16:
```suggestion
  ROOT_PATH = os.path.join(ROOT_PATH, 'resources', 'messages')
```

icubam/messaging/message_formatter.py
```diff
@@ -0,0 +1,35 @@
...
+
+class MessageFormatter:
+  """This class aims at formatting the messages sent to the user.
+
+  Formatting means turning the data in the message into a localized text or
+  html.
+  """
+  ROOT_PATH = os.path.abspath(__file__)
+  for _ in range(3):
+    ROOT_PATH = os.path.dirname(ROOT_PATH)
```
rth, line 14: This is also,
```py
from pathlib import Path

ROOT_PATH = Path(__file__).resolve().parents[2]
```

https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

icubam/messaging/message_formatter.py
olivierteboul: nice

icubam/messaging/message_formatter.py
olivierteboul: nice !

olivierteboul:
Thanks for the review Roman. I believe I have addressed all the comments.
I agree it is a long PR, that has been growing over the week. We still need to wait for PR #378 before merging anyway.

PR #380, opened 2020-04-23
Fix plot legends in flux_dept_lits_dept_visu_4panels

rth:
The issue was that we are using the non default plot style (`seaborn-whitegrid`) which sets `frameon=False` for legends.


rth:
There is still an issue on subplot 2 the twinx plot and y grid can overlap with the legend. I haven't managed to solve it yet with `set_zorder`, and I don't want to spend more time on this (particularly since we may endup not including this subplot in the end). 

I think we should merge as is, it's already an improvement.

PR #382, opened 2020-04-23
SImple script to automate demo deployment, documentation update

### LINKED ISSUE #381: Simple script to automate the installation of a test deployment
Provide a script to automate the deployment of ICUBAM with a test database on a clean VM (with docker/docker-compose support).

- Create default envvar and config files
- Download relevant compose files directly from the ICUBAM repository
- Generate a fake database
- Launch the app's containers

This setup assumes there is an external reverse proxy/ handling redirections and ssl endpoint.

pgraverdy:
Closes #381

**Why?**

Installing a basic instance of the platform for demonstration involves a number of step which are easy to get wrong. A simple, automated procedure is required in order to have people quickly play with the platform with a default database. 

**What?**

This PR adds a simple script that
-  creates the envvar file with default values (its content must be updated by the user and the install script launched again)
- download the relevant config and compose files from the github repository
- create resources folder, and generate the fake databae in it
- launches the app's containers (www, sms, back-office)

The app is then available using the url to access directly the back-office (http://localhost:8890/bo/) and the map (http://localhost:8888/) 

The script assumes docker and docker-compose are installed.

Testing notes
- The script can be downloaded using `wget https://raw.githubusercontent.com/icubam/icubam/pierreg_install_auto/docker/scripts/install_test_deployment.sh`
- while the script relies on compose/config files on the `master` branch, until this branch is merged, the script should be updated to fetch the files from the `pierreg_install_auto` branch (3 instances to replace)
- Since the Docker image cannot be downloaded from Github without credentials, a temporary repository has been setup on Inria's Gitlab. The following tags should be set in the envvar file (after the first execution of the script)
```
IMAGE_NAME=registry.gitlab.inria.fr/praverdy/icubam_docker
IMAGE_TAG=latest
```
- while the script exits a second time to let the user edit the `resources/config.toml` file if required, for testing it can just be re-launched

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -95,9 +95,7 @@ Two kinds of environment variables must be set
 **Docker's environement variables**
 
 - ICUBAM_COMPOSE_CONTEXT: root folder for the build context
-- ICUBAM_CONFIG_PATH: path/filename for the app's configuration file
-- ICUBAM_PROD_DB_PATH: path/filename for the production database
-- ICUBAM_TEST_DB_PATH: path/filename for the test database
+- ICUBAM_RESOURCES_PATH: path/name of the resources folder to mount in the container (e.g., `./resources`). This folder contains, among other things, the app's configuration file and the database).
 - ICUBAM_CERTBOT_PATH: location for the CertBot configuration and result files
 - ICUBAM_NGINX_PATH: location for the Nginx configuration files
 - IMAGE_NAME: name of the Docker image to use
@@ -110,11 +108,14 @@ Files/folders are mounted (bind) in the containers (nginx/certbot) defined in th
 
 The containers expect the following variables to be set in order to launch
 
+- ICUBAM_CONFIG_FILE: filename for the app's configuration file (e.g., `icubam.toml`)
 - SECRET_COOKIE
 - JWT_SECRET
 - GOOGLE_API_KEY
 - TW_KEY
 - TW_API
+- TW_API
+- DB_SALT
 
 These environment variables are not set in the Docker image, but must be set when starting the 
 docker/docker-compose command.  Also check the [install.md](./install.md) documentation for more details.
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -92,35 +92,26 @@ Two kinds of environment variables must be set
 - env. variables used by the app's services
 - env. variables used by docker/docker-compose when building/launching the application 
 
+**Application's environement variables**
+
+Check [Installation instructions](install.md#Configuration) for the required variables
+  
 **Docker's environement variables**
 
 - ICUBAM_COMPOSE_CONTEXT: root folder for the build context
 - ICUBAM_RESOURCES_PATH: path/name of the resources folder to mount in the container (e.g., `./resources`). This folder contains, among other things, the app's configuration file and the database).
+- ICUBAM_CONFIG_FILE: filename for the app's configuration file (e.g., `icubam.toml`) expected in the `./resources` folder
 - ICUBAM_CERTBOT_PATH: location for the CertBot configuration and result files
 - ICUBAM_NGINX_PATH: location for the Nginx configuration files
 - IMAGE_NAME: name of the Docker image to use
 - IMAGE_TAG: tag of the Docker image to use
 - LOGS_DIR: folder where log files (e.g., './logs')
 
-Files/folders are mounted (bind) in the containers (nginx/certbot) defined in the `docker-compose-proxy.yml`file.
-
-**Application's environement variables**
-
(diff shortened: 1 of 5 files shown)


doc/docker.md
```diff
@@ -110,11 +108,14 @@ Files/folders are mounted (bind) in the containers (nginx/certbot) defined in th
...
 The containers expect the following variables to be set in order to launch
 
+- ICUBAM_CONFIG_FILE: filename for the app's configuration file (e.g., `icubam.toml`)
 - SECRET_COOKIE
 - JWT_SECRET
 - GOOGLE_API_KEY
 - TW_KEY
 - TW_API
+- TW_API
+- DB_SALT
```
rth, line 118: Instead of repeating them here, could we link to the required env variables [here](https://github.com/icubam/icubam/blob/master/doc/install.md#configuration) (with `./install.md#configuration` relative link)?

Or at least have some way to list required env variables (excluding docker specific ones) in one place.

docker/build_fake_dbs.sh
```diff
@@ -1,14 +1,5 @@
...
-
-# launch the server
-pwd
-ls -a
+# launch the DB creation
 echo "Build test db"
-PYTHONPATH=. python scripts/populate_db_fake.py --config=resources/config.toml
-
-cp ./test.db ./db/
+PYTHONPATH=. python scripts/populate_db_fake.py --config="/home/icubam/resources/${ICUBAM_CONFIG_FILE:-icubam.toml}"
```
rth, line 5: Just curious: why not install the package instead of modifying PYTHONPATH?

dulacarnold:
I tried running this on a fresh install of ubuntu 18.04 on gcp, and I get the following on. the third run:
```resources/config.toml exist, use iti
-e 
###
Initialize fake database
Building db-builder
ERROR: Cannot locate specified Dockerfile: ./docker/Dockerfile
Removing network gabe_icubam-network
-e 
###
Launch app containers
Creating network "gabe_icubam-network" with driver "bridge"
Building app-server
ERROR: Cannot locate specified Dockerfile: ./docker/Dockerfile```

File: docker/build_fake_dbs.sh (hunk 2 above)
Comment (dulacarnold) [reply docker/build_fake_dbs.sh:5]: This assumes you are running from the root of icubam, but IIUC the goal of the script is to run w/out checking out the git repository no?

Comment (dulacarnold) [reply docker/build_fake_dbs.sh:5]: Sorry, this is from within the image, agreed we should ideally just install icubam to the image instead of importing dynamically.

File: doc/docker.md (hunk 1 above)
Comment (pgraverdy) [reply doc/docker.md:118]: > Is cloning the repo less convenient?

@rth  Here I was just trying to have a solution with the bare minimum (I was hoping for just one file, but then...). And working with branches is then ugly. Indeed, I can just clone and then clean the code after it's launched. Will do that.

Comment (pgraverdy) [reply doc/docker.md:118]: regarding the doc, I agree we should avoid duplication as much as possible to avoid inconsistencies. I'll update.

File: docker/build_fake_dbs.sh (hunk 2 above)
Comment (pgraverdy) [reply docker/build_fake_dbs.sh:5]: Hmm, maybe I missed something but is icubam package available with pip ? (or maybe I'm missing the topics). Anyway, regarding the script above, it should run without the  PYTHONPATH=. already. I'll test and update if that's the case

File: doc/docker.md (hunk 1 above)
Comment (pgraverdy) [reply doc/docker.md]: replaced with relative link

File: docker/build_fake_dbs.sh (hunk 2 above)
Comment (pgraverdy) [reply docker/build_fake_dbs.sh]: Removed PYTHONPATH (not required inside the docker image)

File: docker/docker-compose-init-db.yml [hunk 3]
```diff
@@ -22,9 +23,6 @@ services:
       - type: bind
         source: "${ICUBAM_RESOURCES_PATH:-./resources}"
         target: /home/icubam/resources
-      - type: bind
-        source: "${ICUBAM_COMPOSE_CONTEXT:-.}"
-        target: /home/icubam/db
```
rth, line 27: SO is the DB now mounted as part of the resources folder?

doc/docker.md
```diff
@@ -143,21 +135,20 @@ set -a
 
 **Example .env file**
 
-These work straight out of the box with the docker-compose scripts provided if you run them from the root folder
+For the application's environment variables, check the repository for an up-to-date list.
```
rth, line 138:
```suggestion
For the application's environment variables, see the identically named section above.
```

docker/docker-compose-init-db.yml
pgraverdy: yes

Issue #327, opened 2020-04-18
TypeError: unsupported operand type(s) for *: 'DotMap' and 'int'


rth:
Right now when a config value is missing, it returns a DotMap object. And then errors like 
```
TypeError: unsupported operand type(s) for *: 'DotMap' and 'int'
```
It would be nice to return None by default to make these errors a bit more explicit, and error handling nicer.

I though @dulacarnold mentioned that this was done, but we saw this again during last deployment.

olivierteboul:
It is still one. Thinking about a way to solve it...

rth:
After investigation I don't think we can do anything about it on our side. The problem is that when one does `config.db.non_exitsting_value`, `config.db` is a `DotMap` value and then apparently `DotMap(..).db` returns an empty `DotMap` object. So it's completely up to the dotmap library how to handle that.

rth:
Closing as I don't think we can do much about it.

A possible workaround currently is to check that the key exists e.g. `config.server.has_key(key)` before using it.

PR #383, opened 2020-04-23
Fix config picklability

rth:
Makes `icubam.config.Config` picklable.

This was necessary to be able to start the 3 ICUBAM web-services with in memory config for tests (without using config paths as in https://github.com/icubam/icubam/pull/320). 

Reverts https://github.com/icubam/icubam/pull/320 as it should no longer be necessary.


icubam/config.py
```diff
@@ -47,13 +47,29 @@ def _preprocess(self, conf):
...
-
     return result
 
   def __getitem__(self, key: str):
     if key.upper() == key:
       return self.env.get(key)
-    return self.conf.get(key) if self.conf.has_key(key) else None
+    if self.conf.has_key(key):
+      return self.conf.get(key)
+    raise KeyError(f"key '{key}' does not exist!")
```
rth, line 57: I think the original motivation to return None was to fix https://github.com/icubam/icubam/issues/327 however it doesn't fix it since the issue is in the dotmap library (cf my comment there).

icubam/config.py
```diff
@@ -47,13 +47,29 @@ def _preprocess(self, conf):
...
-    return self[key]
+    raise_err = False
+    try:
+      res = self[key]
+    except KeyError:
+      raise_err = True
+    # Avoid raising an exception inside an exception for traceback
+    # readability.
+    if raise_err:
+      raise AttributeError(f"attribute '{key}' does not exist!")
```
rth, line 68: This is a bit awkward if we reuse `__getitem__` , but anyway the point is that we need to return  `AttributeError` for missing attibutes not None. Otherwise `hasattr(self.config, '__call__')` probably returns True (without it being a callable) and pickle is not able to process this object.

icubam/config.py
```diff
@@ -47,13 +47,29 @@ def _preprocess(self, conf):
...
+    return res
+
+  def __setstate__(self, state):
+    self.__dict__ = state
+
+  def __getstate__(self):
+    return self.__dict__
```
rth, line 75: This is necessary to avoid endless recursions when unpickling for some reason see e.g. https://github.com/tartley/colorama/issues/86#issue-135608263

icubam/config.py
olivierteboul: the problem is not fixed indeed. it is not so easy and would probably need to remove DotMap (or to inherit from it ?)
Because when we do: self.config.a.b.c, it does self.config.a first, this is the responsibility of the Config class. But the getter returns a DotMap object, lets call it dm. Then it is dm.b and it has now nothing to do with the config class any longer, but only has to do with DotMap. Lets' add a TODO for now.

scripts/run_server.py
```diff
@@ -32,17 +36,9 @@ def main(argv):
...
     processes = [
-      mp.Process(
-        target=functools.partial(
-          utils.run_server,
-          cls,
-          config_path=FLAGS.config,
-          env_path=FLAGS.dotenv_path
-        )
-      ) for cls in servers.values()
+      mp.Process(target=set_logging_and_run(cls(cfg, FLAGS.port).run))
```
olivierteboul, line 40: you probably dont want to run all the servers on the same port...

scripts/run_server.py
```diff
@@ -21,6 +20,11 @@
 FLAGS = flags.FLAGS
 
 
+def set_logging_and_run(func):
```
olivierteboul, line 23: this looks like a decorator. but i would not turn it into one.
i think things would be more clear to have a run_one_server that takes a cls and a config:
```
def run_one_server(cls, conf):
  logging.set_verbosity(logging.INFO)
  cls(conf).run()
```

then pass args to the Process (see https://docs.python.org/3.4/library/multiprocessing.html?highlight=process#the-process-class)

icubam/utils.py
```diff
@@ -1,14 +1,6 @@
...
-
-
-def run_server(cls, config_path, env_path):
-  """A helper utility to run Tornado servers from multiprocessing."""
-  cfg = config.Config(config_path, env_path=env_path)
-  logging.set_verbosity(logging.INFO)
-  cls(cfg, None).run()
 
 
 def maybe_init_sentry(cfg, server_name=None):
```
olivierteboul, line 6: could we rename this module sentry.py ? or sentry_utils.py ?

icubam/backoffice/server.py
```diff
@@ -9,7 +9,7 @@
 import tornado.web
 from absl import logging  # noqa: F401
 
-from icubam import base_server, utils
+from icubam import base_server, sentry
```
olivierteboul, line 12: much better

scripts/run_server.py
```diff
@@ -21,6 +20,11 @@
 FLAGS = flags.FLAGS
 
 
+def run_one_server(cls, cfg):
```
olivierteboul, line 23: i would add port=None as argument, this way we can also use this function on line 37

PR #384, opened 2020-04-24
Mechanism for pre-deployment tests

rth:
This adds a few additional tests in `frontend_tests/` to run pre-deployment checks https://github.com/icubam/icubam/issues/311

For now only health checks are included, but other could be added later. We can use selenium if we need JS interaction. However strangely WebDriver API doesn't allow to check for HTTP codes on page load, so using  just `urlopen` here.

There are two usecases,
 1. Start services, then run,
    ```
    pytest --icubam-config=<config> frontend_tests/
    ```
    to check e.g. that proxy is working etc.
 2. Run,
    ```
    pytest --icubam-config=<config>  --run-server frontend_tests/
    ```
    this will start all 3 ICUBAM services in the background, run these checks againt them, then terminate services once tests are done. In this case the config likely needs to be dev config and this would be useful for instance to run these tests in CI.
    We would also likely need to add a `--icubam-config=<envfile_path>` option.

The diff will become significantly smaller once https://github.com/icubam/icubam/pull/383 is merged.


rth:
So the current output with this PR,
```
$ pytest --icubam-config=config.toml frontend_tests/ --run-server -v
[...]
frontend_tests/test_services.py::test_health_urls[server] PASSED                                                                [ 33%]
frontend_tests/test_services.py::test_health_urls[messaging] PASSED                                                             [ 66%]
frontend_tests/test_services.py::test_health_urls[backoffice] PASSED                                                            [100%]
```
(or without `--run-server` if the services are already started). We need to make sure that the tested URLs correctly pass through the proxy for deployment cc @dulacarnold

PR #385, opened 2020-04-24
fixing map for paca

olivierteboul:
- add the possibility to show or not the empty icus
- consistent display for icu without phones
- cut the tree based on level + capacity
- move port of server to 8887 because 8888 is sometimes used by other stuffs

### DIFF
--- a/icubam/icu_tree.py
+++ b/icubam/icu_tree.py
@@ -139,16 +139,29 @@ class ICUTree:
     self._get_leaves(nodes)
     return nodes
 
-  def _extract_below(self, level, nodes):
-    if self.level == level:
-      nodes.append((self, self.get_leaves()))
-    else:
-      for child in self.children.values():
-        child._extract_below(level, nodes)
+  def _extract_below(self, level, nodes, keep_empty=False, max_nodes=None):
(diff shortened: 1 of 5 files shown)


icubam/icu_tree.py
```diff
@@ -139,16 +139,29 @@ def get_leaves(self):
...
-    else:
+  def _extract_below(self, level, nodes, keep_empty=False, max_nodes=None):
+    target = self.LEVELS.index(level)
+    curr = self.LEVELS.index(self.level)
+
+    recurse = curr < target
+    if not recurse:
+      leaves = [n for n in self.get_leaves() if (n.total > 0 or keep_empty)]
+      if max_nodes is not None and len(leaves) > max_nodes:
+        recurse = True
```
rth, line 150: Not sure there is an easy fix, but this (added) line is never run in tests. As it's a somewhat non trivial recursive function it wouldn't hurt to have all code paths testes.

olivierteboul: let me try to figure it out.

olivierteboul: done

PR #388, opened 2020-04-24
Bedcount import endpoint

maximedenes:
This PR makes it possible to import bedcounts CSV files.

Not sure how to handle selection of timezone. @dulacarnold any suggestions? I can probably get the browser timezone through JS, for example.

### DIFF
--- a/icubam/backoffice/handlers/base.py
+++ b/icubam/backoffice/handlers/base.py
@@ -23,7 +23,7 @@ class ObjType(Enum):
       self.REGIONS: "Regions",
       self.USERS: "Users",
       self.TOKENS: "Access Tokens",
-    }.get(self)
+    }[self]
 
 
 class BaseHandler(tornado.web.RequestHandler):
@@ -51,7 +51,7 @@ class BaseHandler(tornado.web.RequestHandler):
     self, data, objtype, create_handler=None, upload=False, **kwargs
   ):
     route = None if create_handler is None else create_handler.ROUTE
-    upload_type = objtype.value if upload else None
+    upload_type = objtype.name if upload else None
     item = data[0] if data else []
     columns = json.dumps([x['key'] for x in item])
     return self.render(
--- a/icubam/backoffice/handlers/base.py
+++ b/icubam/backoffice/handlers/base.py
@@ -2,7 +2,28 @@ import json
 import os.path
 import tornado.locale
 import tornado.web
+from enum import Enum, unique
 from typing import List, Dict, Union, Optional
+
+
+@unique
+class ObjType(Enum):
+  ICUS = 1
+  BEDCOUNTS = 2
+  MESSAGES = 3
+  REGIONS = 4
+  USERS = 5
+  TOKENS = 6
+
+  def __str__(self):
+    return {
+      self.ICUS: "ICUs",
+      self.BEDCOUNTS: "Bed Counts",
+      self.MESSAGES: "Scheduled Messages",
+      self.REGIONS: "Regions",
+      self.USERS: "Users",
+      self.TOKENS: "Access Tokens",
+    }.get(self)
 
 
 class BaseHandler(tornado.web.RequestHandler):
@@ -30,14 +51,14 @@ class BaseHandler(tornado.web.RequestHandler):
     self, data, objtype, create_handler=None, upload=False, **kwargs
   ):
     route = None if create_handler is None else create_handler.ROUTE
-    upload_type = objtype if upload else None
+    upload_type = objtype.value if upload else None
     item = data[0] if data else []
     columns = json.dumps([x['key'] for x in item])
     return self.render(
       "list.html",
       data=data,
       columns=columns,
-      objtype=objtype,
+      objtype=str(objtype),
       create_route=route,
       upload_type=upload_type,
       **kwargs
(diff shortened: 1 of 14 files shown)


rth:
> Not sure how to handle selection of timezone. 

Well ideally timezone should be indicated together with the timestamp columns inside the CSV. If it's not possible I imagine we could also add a timezone argument on the backend and maybe yes query [it from the browser in JS](https://stackoverflow.com/a/1091399/1791279)?

icubam/backoffice/handlers/base.py
```diff
@@ -30,7 +30,7 @@ def render_list(
     self, data, objtype, create_handler=None, upload=False, **kwargs
   ):
     route = None if create_handler is None else create_handler.ROUTE
-    upload_type = route if upload else None
+    upload_type = objtype if upload else None
```
olivierteboul, line 33: i don t understand what motivated this change.. can you give an example?

icubam/backoffice/handlers/upload.py
```diff
@@ -29,9 +29,9 @@ def post(self) -> None:
 
     sync = synchronizer.CSVSynchronizer(self.db)
     sync_fns: Dict[str, Callable[..., int]] = {
-      'user': sync.sync_users_from_csv,
-      'icu': sync.sync_icus_from_csv,
-      'bedcounts': sync.sync_bedcounts_from_csv
+      'Users': sync.sync_users_from_csv,
```
olivierteboul, line 32: i don t like too much this dependence here. it might not have been optimal either, but this `objtype` is really used to show something to the user, be translated, etc. I would rather use an enum so as to make sure things are consistent and do not depend on variables meant to be printed/displayed.

maximedenes: I switched to an enum, thanks.

icubam/backoffice/handlers/base.py
maximedenes: The motivation was in the corresponding commit message. This confusion between route and object type made it impossible to upload data which does not have a create_handler.

maximedenes:
@rth Thanks for the review, I added a test.

For timezones, I simply switched the CSV format to timezone-aware datetimes (according to ISO standard), so we can read everything from one field.

icubam/backoffice/server_test.py
```diff
@@ -90,3 +90,32 @@ def test_consent_reset(self):
...
     self.assertIsNone(resp_data['error'])
+
+  def test_upload(self):
+    handler = upload.UploadHandler
+    with mock.patch.object(base.BaseHandler, 'get_current_user') as m:
+      m.return_value = self.admin
+      data_icus = dict()
+      with open("resources/test/icu2.csv") as csv_f:
+        data_icus['data'] = csv_f.read()
+        data_icus['objtype'] = base.ObjType.ICUS.value
```
rth, line 101: Using Enum internally sounds good, but for API users I think it's still nicer and more expressive to pass a string as a argument instead of int (and checking out the mapping table). For instance here we could have used enum name instad.

icubam/backoffice/handlers/upload.py
```diff
@@ -28,13 +28,13 @@ def post(self) -> None:
...
-      'icu': sync.sync_icus_from_csv,
-      'bedcounts': sync.sync_bedcounts_from_csv
+    sync_fns: Dict[int, Callable[..., int]] = {
+      base.ObjType.USERS.value: sync.sync_users_from_csv,
+      base.ObjType.ICUS.value: sync.sync_icus_from_csv,
+      base.ObjType.BEDCOUNTS.value: sync.sync_bedcounts_from_csv
     }
     objtype = data.get('objtype', None)
-    sync_fn = sync_fns.get(objtype, None)
+    sync_fn = sync_fns.get(int(objtype), None)
```
rth, line 37: If we pass the enum name in the API, I think here we could do, 
```suggestion
    objtype_name = data.get('objtype', None)
    objtype = base.ObjType[objtype_name]
    sync_fn = sync_fns.get(objtype, None)
```
and use `base.ObjType.USERS` etc. as hash keys directly above.

resources/test/bedcounts.csv
```diff
@@ -1,2 +1,2 @@
 icu_name,n_covid_occ,n_covid_free,n_ncovid_occ,n_ncovid_free,n_covid_deaths,n_covid_healed,n_covid_refused,n_covid_transfered,create_date
-hopital3,11,12,13,14,15,16,17,18,2020-04-16 23:00
\ No newline at end of file
+hopital3,11,12,13,14,15,16,17,18,2020-04-16 23:00+02
```
rth, line 2: We control the format of CSV right? So we add the timezone in the csv as needed?

scripts/csv_import.py
```diff
@@ -19,9 +19,6 @@
 flags.DEFINE_string(
   "bedcounts_csv", None, "Path to csv file containing bedcounts data."
 )
-flags.DEFINE_string(
-  "timezone", "UTC", "Timezone in which dates should be interpreted."
-)
```
rth, line 24: I don't use this script so I don't really have an opinion about it.

resources/test/bedcounts.csv
maximedenes: Yes, AFAIK nobody relies on the format yet.

scripts/csv_import.py
maximedenes: It should be ok, I introduced this option recently.

maximedenes:
@rth I switched to using enum names, thanks for the suggestion.

PR #389, opened 2020-04-24
Icubam-specific data analysis.

dulacarnold:


icubam/analytics/preprocessing.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+def format_data(d):
+  d["datetime"] = pd.to_datetime(d["date"])
+  d["date"] = d["datetime"].dt.date
+  d = d[ALL_COLUMNS]
+  return d
+
+
+def preprocess_data(
```
rth, line 40: So this is preprocesing mostly unchanged but with additional comments from the predicu file, right?

dulacarnold: Remove as much of FR-specific processing, more comments, and only ICUBAM data.

rth:
I think we should not try to add all new plots here, but merge it in a mostly working state and iterate. Let me know if I can push a few commits to fix CI issues. Du we want to remove predicu/ in this PR as well?

icubam/analytics/data.py
```diff
@@ -0,0 +1,35 @@
...
+  "icu_name", "date", "datetime", "department", "region_id", "region",
+  "create_date"
+] + CUM_COLUMNS + NCUM_COLUMNS)
+
+
+def load_bed_counts(db, preprocess: bool = False):
+  """Load Bedcount data from ICUBAM DB"""
+  bc = store.to_pandas(db.get_bed_counts(), max_depth=2)
+  if preprocess:
+    bc = preprocessing.preprocess_bedcounts(bc)
```
rth, line 33: I think it's better to explicitly apply processioning function after the load. Among other things because there are multiple way to do preprocessing with potentially different parameters. Users of this module should read `preprocess_bedcounts` docstring instead of just applying `preprocess=True` here which obfuscates it.

icubam/analytics/plots/__init__.py
```diff
@@ -83,8 +69,9 @@ def plot_int(
   ls="solid",
   s=3,
   fill_below=False,
+  kind="quadratic"
 ):
-  f = scipy.interpolate.interp1d(x, y, kind="quadratic")
+  f = scipy.interpolate.interp1d(x, y, kind=kind)
```
rth, line 74: I haven't noticed it's quadratic interpolation. This can potentially overfit a lot.

icubam/analytics/preprocessing.py
```diff
@@ -1,15 +1,10 @@
...
-from typing import Optional
 
 import numpy as np
 import pandas as pd
 
-from icubam.predicu.data import (
-  ALL_COLUMNS, BEDCOUNT_COLUMNS, CUM_COLUMNS, DATA_PATHS, NCUM_COLUMNS,
-  format_data
-)
+import icubam.analytics.data as data
```
rth, line 7: data is a really bad module name. For instance, `data.CUM_COLUMNS` in a function that processes some data makes me thing that `data` is a dataframe. I think the previous imports were clearer.

scripts/gen_plots.py
```diff
@@ -0,0 +1,47 @@
...
+from absl import app, flags
+
+from icubam import config
+from icubam.db.store import create_store_factory_for_sqlite_db
+from icubam.analytics import data, plots
+
+flags.DEFINE_string('config', config.DEFAULT_CONFIG_PATH, 'Config file.')
+flags.DEFINE_string('output_dir', '/tmp', 'Output dir for plots.')
+flags.DEFINE_string('plot_name', 'base_dashboard_plots', 'Plot name.')
+FLAGS = flags.FLAGS
```
rth, line 15: We kind of have already this script in `analytics.__main__` done with click which has a significantly better documentation.

dulacarnold: Yeah, but since we are using dedicated .py files for all other scripts it seemed a bit non-standard to use the __main__ trick only here.  Happy to reconsider.

icubam/analytics/preprocessing.py
dulacarnold: Good point, we can refactor to another name, but IIRC data was the original module name...  I'm not super keen on direct imports into the local namespace as it gets hard to follow what comes from where when you're not super familiar with the code however.  Do you have a better name suggestion?

icubam/analytics/plots/__init__.py
dulacarnold: Yes, we can raise an error if it fails maybe?

scripts/gen_plots.py
rth: If they are part of the main module and use `click` library, we could have auto-generated the docs for them see e.g. https://paris-saclay-cds.github.io/ramp-docs/ramp-board/stable/cli.html

Otherwise in the current situation the only way for an external contributor to discover what they do is to run `script/<name>.py --help` one by one or read the source code.

icubam/analytics/preprocessing.py
rth: Maybe `datasets.py`?

icubam/analytics/data.py
dulacarnold: Done

icubam/analytics/preprocessing.py
dulacarnold: Done

scripts/gen_plots.py
dulacarnold: Agreed, but I'm not sure that's worth the inconsistency relative to our other scripts.  @olivierteboul do you have an opinion?

icubam/analytics/plots/__init__.py
```diff
@@ -0,0 +1,184 @@
+import itertools
```
olivierteboul, line 1: why does it have to be in __init__.py and not a standard module?

scripts/gen_plots.py
olivierteboul: I would keep it like this for consistency with other ways to run scripts.

rth:
Thanks for working on this, reviewing now. One thing that concerns me is that coverage decreased with respect to master, meaning that we are not testing plots that were previously tested.

PR #392, opened 2020-04-28
TST More backoffice handler tests

rth:
Ref: https://github.com/icubam/icubam/issues/197

Additional tests for the POST method of backoffice region, icus, tokens, users endpoints.

**Edit:** Overall this brings the code coverage of backoffice handlers from 62% to 77% in this PR.


icubam/backoffice/server_test.py
```diff
@@ -119,3 +120,95 @@ def test_upload(self):
...
+
+      data_email = 'test@test.org'
+      data = {
+        'user_id': '',
+        'name': 'TestUser',
+        'telephone': '00000000000',
+        'is_active': 'on',
+        'email': "test%40test.org",
+        'password': 'test123',
+        'icus%5B%5D': '1',
```
rth, line 203: `%5B%5D` are `[` and `]` likely to indicate this value as a list. Not sure how this works exactly but in any case that's the value that gets send to the server during a POST request currently.

olivierteboul: Yes it is the way the post requests send a list and not a single value (the first one).
By the way, i would prefere to use `tornado.escape.url_escape('icus[]')` instead, or rather define your data dict, and then `{tornado.escupe.url_escape(k): tornado.escupe.url_escape(v) for (k, v) in data.items()}`

olivierteboul: or below when calling self.fetch

rth: Yes, I considered escaping as well. Will do. Thanks for the review!

rth: Actually, `urllib.parse.urlencode` can both convert dict to query and escape.

PR #393, opened 2020-04-28
FIX commit changes in wipe script

rprimet:
Seems that there was a missing `commit()` in the wipe script. This wasn't found during tests so this PR  probably requires at least a regression test...


rprimet:
The end of the function uses the `Store` object rather than the underlying session so that handles the commits appropriately I would guess. 

I'm not sure how to test, and I'm unsure whether the unit tests succeeded because the test db was in-memory or because the reads (to confirm the value changes) that came after the updates would issue an implicit commit?

We might as well merge as-is and I'll try to write a regression test afterwards?

PR #397, opened 2020-04-28
Disclaimer link on maps and its page

VincentRouvreau:


icubam/www/handlers/home.py
```diff
@@ -40,6 +53,16 @@ def initialize(self, config, db_factory):
     super().initialize(config, db_factory)
     self.regions = None
 
+  def get_disclaimer_url(self):
+    if self.config.server.has_key('disclaimer'):
+      path = self.config.server.disclaimer
+      # The user has already agreed, we skip.
+      if path is not None and os.path.exists(path):
+        return "<a href='{}disclaimer'>disclaimer</a>".format(
+          self.config.server.base_url
```
rth, line 62: I don't think it should contain the `base_url`. For instance I start the dev server on a remote url for testing, and overall . This however redirects to localhost (for a dev server) and fails to load.

icubam/www/handlers/home.py
```diff
@@ -40,6 +53,16 @@ def initialize(self, config, db_factory):
     super().initialize(config, db_factory)
     self.regions = None
 
+  def get_disclaimer_url(self):
+    if self.config.server.has_key('disclaimer'):
+      path = self.config.server.disclaimer
+      # The user has already agreed, we skip.
```
rth, line 59: I don't understand this comment. This only checks paths existence not user agreement.

icubam/www/handlers/home.py
```diff
@@ -40,6 +53,16 @@ def initialize(self, config, db_factory):
...
+      if path is not None and os.path.exists(path):
```
rth, line 60: Can you actually put `None` as a value in toml without it being interpreted as a string? Just curious.

icubam/www/handlers/disclaimer.py
```diff
@@ -0,0 +1,29 @@
...
+  def initialize(self, config, db_factory):
+    super().initialize(config, db_factory)
+
+  def get_disclaimer_html(self):
+    """To show a disclaimer page if specified in configuration."""
+    path = self.config.server.disclaimer
+    # The user has already agreed, we skip.
+    if path is not None and os.path.exists(path):
+      with open(path, 'r') as fp:
+        return fp.read()
```
rth, line 18: Instead of opening the file here, can we just `{% include *file_path* %}` in the template? I guess that might also use the template caching (not sure) https://www.tornadoweb.org/en/stable/guide/templates.html#configuring-templates

icubam/www/templates/disclaimer.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+
+{% block links %}
+{% end %}
+
+{% block content %}
+<div class="modal-body">
+  <p>{% raw disclaimer %}</p>
```
rth, line 8: Maybe the `<p>` should be in the disclaimer HTML? Otherwise it can't contain several paragraphs.

icubam/www/handlers/disclaimer.py
rth: Also if path is in the config but it not valid, I feel we should at least raise a warning to indicate that something is wrong.

icubam/www/handlers/home.py
VincentRouvreau: Bad copy/paste comment. Fixed on 2769ea7

icubam/www/handlers/home.py
VincentRouvreau: Bad copy/paste comment. Fixed on 2769ea7

icubam/www/handlers/home.py
VincentRouvreau: You are right. If I set a value to `None` in the toml file, I am facing a `TomlDecodeError`. Removed this test on b2ad56e

icubam/www/templates/disclaimer.html
VincentRouvreau: Remove <p> tag on 9e9e253

VincentRouvreau:
> Could you also please add a test in https://github.com/icubam/icubam/blob/master/icubam/www/server_test.py to make a GET call on this endpoint? Thanks!

I added a test on fake disclaimer on 9dd24db 

> Also if path is in the config but it not valid, I feel we should at least raise a warning to indicate that something is wrong.

You are right this is better handled like this 962374c

icubam/www/handlers/disclaimer.py
VincentRouvreau: > Instead of opening the file here, can we just `{% include *file_path* %}` in the template? I guess that might also use the template caching (not sure) https://www.tornadoweb.org/en/stable/guide/templates.html#configuring-templates

If I use `{% include '%s' % disclaimer %}` with `disclaimer` as the file path, I am facing:
`FileNotFoundError: [Errno 2] No such file or directory: "/path/to/icubam/icubam/www/templates/%s' % disclaimer"`
This is because include is generally used to include a template in another one. This means I would have to add another template directory with the one where disclaimer.html is.
I think the `{% raw disclaimer %}` solution is better, no ? But I can understand that if the html is big it is not a nice solution.

VincentRouvreau:
> @VincentRouvreau Any chance you would be available to respond to comment so we can include this in the next release?

Sorry about that, I had other things on the stove. Thanks for the review, I am done.

icubam/www/handlers/disclaimer.py
rth: Fair enough. Thanks for checking.

icubam/www/handlers/home.py
```diff
@@ -40,6 +50,15 @@ def initialize(self, config, db_factory):
     super().initialize(config, db_factory)
     self.regions = None
 
+  def get_disclaimer_url(self):
+    """To show a disclaimer link if defined in configuration."""
+    if self.config.server.has_key('disclaimer'):
+      return "<a href='{}disclaimer'>disclaimer</a>".format(
+        self.config.server.base_url
```
rth, line 57:
```suggestion
      return "<a href='/disclaimer'>disclaimer</a>"
```

because it otherwise it would not allow accessing this URL for the dev server from another location than localhost.

icubam/www/handlers/home.py
```diff
@@ -14,6 +14,15 @@ def initialize(self, config, db_factory):
     super().initialize(config, db_factory)
     self.icu = None  # should be overwritten by the authentication
 
+  def get_disclaimer_url(self):
+    """To show a disclaimer link if defined in configuration."""
+    if self.config.server.has_key('disclaimer'):
+      return "<a href='{}disclaimer'>disclaimer</a>".format(
+        self.config.server.base_url
```
rth, line 21:
```suggestion
      return "<a href='/disclaimer'>disclaimer</a>"
```

PR #399, opened 2020-04-29
Re-add plot tests for analytics module

rth:
Re-adds some of the predicu tests for plots that didn't get included in https://github.com/icubam/icubam/pull/389
(otherwise plots are currently untested).

Also test the `spread_cum_jump` option in `preprocess_bedcounts`.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,7 +5,7 @@ coverage:
     project:
       default:
         # Commits pushed to master should not make the overall
-        # project coverage decrease by more than 4%:
+        # project coverage decrease by more than this value:
         target: auto
         threshold: 4%
     patch:
@@ -18,4 +18,4 @@ coverage:
         # extension:
         # https://github.com/codecov/browser-extension
         target: auto
-        threshold: 50%
+        threshold: 15%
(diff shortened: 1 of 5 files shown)


icubam/analytics/preprocessing.py
```diff
@@ -302,8 +303,14 @@ def spread_cum_jumps(d, icu_to_first_input_date):
       for col in cols:
         if col in already_fixed_col:
           continue
-        beg_val = dg.loc[dg.date == beg, col].values[0]
-        end_val = dg.loc[dg.date == end, col].values[0]
+        beg_val = dg.loc[dg.date == beg, col]
+        if not len(beg_val):
+          continue
```
rth, line 308: Otherwise this fails with key error on test data, particularly due to the hardcoded `date_begin_transfered_refused` above.

.codecov.yml
```diff
@@ -18,4 +18,4 @@ coverage:
         # extension:
         # https://github.com/codecov/browser-extension
         target: auto
-        threshold: 50%
+        threshold: 15%
```
rth, line 21: This means roughly that new PR need to have at least ~80% (auto) - 15%  = 65% code coverage.

PR #400, opened 2020-04-30
Add Docker user envvar

pgraverdy:
**Why**

When creating the default/test database (or other configuration files) within a container, the default user is root.  On Linux, this is an issue when these files are created on a mounted folder.
In particular, on CI, this raise error messages when cleaning up files.

**What**

Add the setup in the docker-compose file, and information in the documentation.
 
**Remaining**

- Add info in documentation


dulacarnold:
What value will this get populated by then?  The default VM user?

pgraverdy:
@dulacarnold  By default, the default user on Linux is 0:0 (root). It's a bit different on MacOSX since Docker is a native app. I updated the doc to document how to setup the values so that the files (e.g., db, nginx/certbot) are created with the users uid/gid.

PR #410, opened 2020-05-04
Docker: single entry point?

olivierteboul:
This PR is more like a suggestion since I have not tested it.

However, looking at the code, i believe we do not need one bash script per service, knowing that they all call` python3 -m run.server --config=....`, only with a different `--server=` parameter. 

I believe having less file is better than more. Let me know what you think about it.


pgraverdy:
Indeed, it amkes for a cleaner repo. The main thing to check is that env. variables are properly passed on to the python script (sometimes it's tricky with Docker).

docker/Dockerfile
```diff
@@ -48,10 +48,9 @@ RUN rm -rf ~/.cache/pip
...
-# expose ports for flask app - both server and backoffice
+# expose ports for tornado app - both server and backoffice
 EXPOSE 8888 8890
 
-# default commend when entering the container
-ENTRYPOINT ["/bin/bash", "-c"]
-# default command
-CMD ["./docker/start_server_www.sh"]
+# default command when entering the container
+ENTRYPOINT ["python3", "-m", "scripts.run_server", "--config", "${ICUBAM_CONFIG_FILE:-icubam.toml}"]
```
VincentRouvreau, line 55: I understand the simplification, but sometimes we also need to enter the docker image to run some scripts in maintenance (scripts/wipe_sensitive_info_db.py, scripts/populate_fake_db.py, scripts/get_url_id.py, ...).
I am not a docker expert, but here we will have to set `--entrypoint` to be able to use it, that's it ?

VincentRouvreau:
We all agree it would be nice if we could remove the `./docker/start_server_*.sh` scripts. This PR is just a suggestion, so I will let @pgraverdy say if it is possible or not.

docker/Dockerfile
olivierteboul: so if i remember correctly, you can always enter the docker container via `docker exec -it <container_id> /bin/bash`

olivierteboul:
thanks! This PR is only a suggestion indeed.

pgraverdy:
@olivierteboul 

In the docker-compose files, if the full command is set for each service in exec mode like this
```
    command: ["python3", "scripts/run_server.py", "--server=backoffice", "--config=/home/icubam/resources/${ICUBAM_CONFIG_FILE:-icubam.toml}"]
```
then we can delete the extra shell scripts, all is fine wrt environment variable (need to test also in shell mode). We can also remove the entrypoint/cmd in the dockerfile.

Alternatively, we can have the entrypoint in the dockerfile with the common parameters, for example
```
ENTRYPOINT: ["python3",  " "scripts/run_server.py" ...] 
```
and have the extra parameters in the command part of the docker-compose.
```
    command:  --server=backoffice
```

Second solution may be cleaner but then it splits the command in two different files so more difficult to understand ?

What do you think ?

PR #411, opened 2020-05-13
Faster /version endpoint

rth:
The /version endpoint is currently extremely slow (2s locally, 10s on prod server) because a query of all bedcounts is done as part of health checks (and that grows slower with more data).

Querying only latest bedcounts for each ICU is much faster (30ms locally). Assuming we still want to do some queries in this endpoint as opposed to just returning the version.


VincentRouvreau:
This is maybe what explains the status on the supervision server:
image

rth:
Thanks @VincentRouvreau !

PR #417, opened 2020-05-27
Add sphinx documentation

rth:
This adds a sphinx documentation of the markdown files in the `doc/` folder. It makes navigating it a bit easier, as there is a table of contents etc. We can also potentially auto-generate some docs, e.g. for scripts.

A preview of the rendered version can be seen at https://test-doc-icubam.readthedocs.io/en/latest/index.html There is quite a few things that can be improved in the docs (there are some outdated parts), but I'd like to do that in a separate PR after this one,

To make it work, we need to,
 - create a pyodide project at https://readthedocs.org/ for the `icubam/icubam` repo (I don't have enough permissions to do it, currently using my fork for the demo). Add me as a co-maintainer (login: rth) and I'll setup the rest, if needed, though it should mostly work out of the box.
 - (optional) by default the docs would be at `icubam.readthedocs.io` to use the `icubam.net` domain we would need to setup a CNAME from e.g. `docs.icubam.net` to `readthedocs.io` (https://docs.readthedocs.io/en/stable/custom_domains.html)


doc/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
+
+pushd %~dp0
+
+REM Command file for Sphinx documentation
+
+if "%SPHINXBUILD%" == "" (
+	set SPHINXBUILD=sphinx-build
+)
+set SOURCEDIR=.
```
rth, line 10: The makefiles are auto-generated with sphinx, and don't need to be reviewed.

rth:
https://icubam.readthedocs.io/en/latest/  is now working.

PR #419, opened 2020-05-29
Add Web API documentation and update release notes 

rth:
A constraint with the API docs is that they are not updated automatically (e.g. with OpenAPI etc) and so have to be manually updated each time there is a change in handlers. Since we don't plan much changes in the future it should be OK I think.

**Edit:** rendered version in https://test-doc-icubam.readthedocs.io/en/web-api/api.html


rth:
Thanks for the review @dulacarnold !

PR #425, opened 2020-06-09
Delete build-events.yml

rprimet:
This file was used to trigger a deployment on the systemd staging instance. As this instance is being retired, this deployment step is no longer needed (and if/when a new deployment pipeline is needed, it might be better to not have it depend on a hook in the main repository and adopt a "pull" approach perhaps?)


rth (approved):
Sounds good, thanks!

PR #427, opened 2020-11-03
Proposal to Antoine request to add a view with the sum of all beds (COVID+ + COVID-)

epoiseau:


icubam/icu_tree.py
```diff
@@ -2,166 +2,176 @@
...
-      if max_nodes is not None and len(leaves) > max_nodes:
-        recurse = True
-      else:
-        nodes.append((self, leaves))
-
-    if recurse:
-      for child in self.children.values():
-        child._extract_below(
-          level, nodes, keep_empty=keep_empty, max_nodes=max_nodes
+    LEVELS = ['country', 'region', 'dept', 'city', 'icu']
```
rth: Is it necessary to change the indentation level in this PR? I'm also not a bit fan of 2 space indents, but at least it should be consistent across the project IMO (either 2 or 4 everywhere). It makes it a bit harder to review this PR.. 

**Edit:** also this should actually be enforced by the yapf formatter, which explains the failure.

icubam/map_builder.py
```diff
@@ -76,7 +76,7 @@ def prepare_jsons(
       'lat': center_icu.lat,
       'lng': center_icu.long
     } if center_icu else None
-    for covid in [True, False]:
+    for covid in [True, False, None]:
```
rth, line 83: It would help to have a comment (here or in another file) explaining what is covid=None.

rth (commented):
Thanks! A couple of comments..

epoiseau (commented):
Roman. I have update the code with comments and configured my IDE to use the right number of spaces

